PR #116, opened 2020-04-27
Add db script to clean up scheduled calls that are at least 24 hours old

jivdhaliwal:
# What
Add a db script to clean up scheduled calls that are at least 24 hours old

# Why
Allows us to easily run this cleanup job on all envs

# Screenshots

# Notes
Run with `npm run cleandb`


db/scripts/cleanup_scheduled_calls.js
```diff
@@ -0,0 +1,27 @@
...
+  const pgp = require("pg-promise")({});
+  const connectionURL = process.env.URI;
+  return pgp({
+    connectionString: connectionURL,
+    ssl: { rejectUnauthorized: false },
+  });
+}
+
+async function cleanupScheduledCalls() {
+  const db = await getDb();
```
antonyoneill, line 14: I reckon you could use the `AppContainer` here, rather than building the connection above

db/scripts/cleanup_scheduled_calls.js
```diff
@@ -0,0 +1,27 @@
+async function getDb() {
+  const dotenv = require("dotenv");
```
antonyoneill, line 2: If you did use the AppContainer you'd still need to import the dotenv library.

Newer JS syntax for this would be 
```javascript 
import dotenv from 'dotenv';

dotenv.config();
```

jivdhaliwal: This results in `SyntaxError: Cannot use import statement outside a module`. Most likely because the script is being run as is and not transpiled

db/scripts/cleanup_scheduled_calls.js
jivdhaliwal: This results in `SyntaxError: Cannot use import statement outside a module`. Most likely because the script is being run as is and not transpiled

PR #119, opened 2020-04-27
Update start and end time of Whereby meetings

chubberlisk:
# What

- Update start time of Whereby meetings to the scheduled time
- Update end time of Whereby meetings to 1 year after the scheduled time

# Why

From user testing, we found out that visits considerably run over. So we want to increase the time that a meeting can last.

# Screenshots

N/A

# Notes

Wasn't able to double check that Whereby still works, if you can test locally might be worth it.


chubberlisk:
Tested using the Heroku preview app and twas 🆗 .

PR #120, opened 2020-04-27
Add ability for pulling data to rebook a visit

steveknightmadetech:
# What
Add an additional parameter to schedule-visit page to allow rebooking of visit

# Why
Need ability to rebook a visit when a visit is complete

# Screenshots

# Notes

### DIFF
--- a/pages/wards/[id]/schedule-visit.js
+++ b/pages/wards/[id]/schedule-visit.js
@@ -285,9 +285,9 @@ export const getServerSideProps = propsWithContainer(
           initialContactNumber: contactNumber,
           initialCallDateTime: callDateTime,
         };
-      } else if (query.rebook) {
+      } else if (query.rebookCallId) {
         const { scheduledCall, error } = await retrieveVisitByCallId(container)(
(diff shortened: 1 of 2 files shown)


pages/wards/[id]/schedule-visit.js
```diff
@@ -258,36 +260,48 @@ const queryContainsInitialData = (query) => {
...
+        props = {
+          ...props,
+          initialPatientName: patientName,
+          initialContactName: contactName,
+          initialContactNumber: contactNumber,
+          initialCallDateTime: callDateTime,
+        };
+      } else if (query.rebook) {
+        const { scheduledCall, error } = await retrieveVisitByCallId(container)(
+          query.rebook
```
Comment (jivdhaliwal) [inline pages/wards/[id]/schedule-visit.js:290]: What do we think about renaming this to rebookCallId ? Otherwise the only way you know that this is actually a callId being passed through is by it being used in `retrieveVisitByCallId`

Comment (steveknightmadetech) [reply pages/wards/[id]/schedule-visit.js]: Great idea - done :)

PR #127, opened 2020-04-28
usecase to delete a visit by callid

steveknightmadetech:
# What
Use case to delete visit row from database table scheduled_call_table

# Why
To allow deletion of scheduled calls

# Screenshots

# Notes

### DIFF
--- a/src/usecases/deleteVisitByCallId.js
+++ b/src/usecases/deleteVisitByCallId.js
@@ -6,20 +6,14 @@ const deleteVisitByCallId = ({ getDb }) => async (callId) => {
       `DELETE FROM scheduled_calls_table WHERE call_id = $1 LIMIT 1`,
       callId
     );
-
-    const result = results[0];
-    console.log("result", result);
-
     return {
-      result: {
-        success: true,
-      },
+      success: true,
       error: null,
     };
   } catch (error) {
     console.error(error);
     return {
-      result: null,
+      success: false,
       error: error.toString(),
     };
   }
--- a/src/usecases/deleteVisitByCallId.js
+++ b/src/usecases/deleteVisitByCallId.js
@@ -1,0 +1,28 @@
+const deleteVisitByCallId = ({ getDb }) => async (callId) => {
+  const db = await getDb();
+  console.log("deleting visit for  ", callId);
+  try {
+    const results = await db.any(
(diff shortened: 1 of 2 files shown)


src/usecases/deleteVisitByCallId.js
```diff
@@ -0,0 +1,28 @@
...
+
+    const result = results[0];
+    console.log("result", result);
+
+    return {
+      result: {
+        success: true,
+      },
+      error: null,
+    };
```
chubberlisk, line 18: What do you think of the below? I don't think the `result` key brings much value and simplifies the check that the usecase has successfully deleted a visit.

```suggestion
    return {
      success: true,
      error: null,
    };
```

src/usecases/deleteVisitByCallId.js
```diff
@@ -0,0 +1,28 @@
...
+  } catch (error) {
+    console.error(error);
+    return {
+      result: null,
+      error: error.toString(),
+    };
```
chubberlisk, line 24: If the above suggestion is the way we want to go, then we can updated this to `{ success: false, error: error.toString() }`.

src/usecases/deleteVisitByCallId.test.js
```diff
@@ -0,0 +1,41 @@
...
+        return {
+          any: jest.fn(() => {
+            throw new Error("DB Error!");
+          }),
+        };
+      },
+    };
+
+    const { error } = await deleteVisitByCallId(container)("cb238rfv23cuv3");
+    expect(error).toBeDefined();
```
chubberlisk, line 39: In the code I **_think_** we return the error message (guessing `error.toString()` does that, not sure), so maybe we can do an expectation that we return that instead of checking if it's been defined.

```suggestion
    expect(error).toEqual("DB Error!");
```

steveknightmadetech: Agree :)

src/usecases/deleteVisitByCallId.js
steveknightmadetech: Agree :)

src/usecases/deleteVisitByCallId.js
steveknightmadetech: Agree :)

PR #136, opened 2020-04-28
Add ward id to scheduled calls

jivdhaliwal:
# What
Adds a ward_id foreign key to the scheduled calls table

# Why
This will eventually allow for support of multiple wards

# Screenshots

# Notes
Allowing for NULL ward_ids for now as we will need to seed the environments with ward records and then update the scheduled calls login to use the stored ward

### DIFF
--- a/db/migrations/005_add_ward_id_to_scheduled_calls_table.sql
+++ b/db/migrations/005_add_ward_id_to_scheduled_calls_table.sql
@@ -1,0 +1,1 @@
+ALTER TABLE scheduled_calls_table ADD ward_id int REFERENCES wards (id);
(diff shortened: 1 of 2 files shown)


db/migrations/005_add_ward_id_to_scheduled_calls_table.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE scheduled_calls_table ADD ward_id int REFERENCES wards (id);
```
jbevan4, line 1: A lot of the documentation has INTEGER over int. Do we use int elsewhere?

jivdhaliwal: int is simply an alias for integer (see https://www.postgresql.org/docs/current/datatype.html#DATATYPE-TABLE)

jbevan4: coolio 👍

PR #139, opened 2020-04-29
Associate scheduled calls and wards

jivdhaliwal:
# What
Associated scheduled calls with the Ward stored in the database

# Why
Work towards supporting multiple wards

# Screenshots

# Notes
For now we are assuming there is only one ward in the database (until we add the ability
to login as different wards).

### DIFF
--- a/src/usecases/retrieveWardById.js
+++ b/src/usecases/retrieveWardById.js
@@ -1,0 +1,27 @@
+const retrieveWardById = ({ getDb }) => async (wardId) => {
+  const db = await getDb();
+  console.log("Retrieving ward for  ", wardId);
+  try {
+    // Assuming there is only one Ward to choose from for now
+    const ward = await db.oneOrNone("SELECT * FROM wards LIMIT 1", wardId);
+
+    console.log("Ward: ", ward);
+
+    return {
+      ward: {
+        id: ward.id,
+        name: ward.name,
+        hospitalName: ward.hospital_name,
+      },
+      error: null,
+    };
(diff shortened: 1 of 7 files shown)


src/usecases/retrieveWardById.js
```diff
@@ -0,0 +1,32 @@
+const retrieveWardById = ({ getDb }) => async (wardId) => {
+  const db = await getDb();
+  console.log("Retrieving ward for  ", wardId);
+  try {
+    // const ward = await db.oneOrNone(
+    //   "SELECT * FROM wards WHERE id = $1 LIMIT 1",
+    //   wardId
+    // );
```
jbevan4, line 8: Should we remove this comment if it's not needed?

db/migrations/005_add_ward_id_to_scheduled_calls_table.sql
```diff
@@ -1 +1 @@
-ALTER TABLE scheduled_calls_table ADD ward_id int REFERENCES wards (id);
+ALTER TABLE scheduled_calls_table ADD ward_id INTEGER REFERENCES wards (id);
```
antonyoneill, line 1: I know `int` and `INTEGER` are equivalent, but I'm very wary of making any modifications to migrations that have already been run on production, as this will not be run again.

Can you revert please?

src/usecases/retrieveWardById.js
jbevan4: tldr; saw your slack post

PR #142, opened 2020-04-29
Use ward id from token

antonyoneill:
# What
Refactor verifyToken to provide the getServerSideProps with the authenticationToken if valid.
Use auth token to retrieve ward ID

# Why
Prepare for removing ward id from url path

# Screenshots

# Notes

### DIFF
--- a/pages/wards/[id]/schedule-visit.js
+++ b/pages/wards/[id]/schedule-visit.js
@@ -253,8 +253,8 @@ const queryContainsInitialData = (query) => {
 
 export const getServerSideProps = propsWithContainer(
   verifyToken(
-    async ({ query, container }) => {
-      const { id } = query;
+    async ({ query, container, authenticationToken }) => {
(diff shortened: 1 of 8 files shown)


pages/wards/[id]/schedule-visit.js
```diff
@@ -253,8 +253,8 @@ const queryContainsInitialData = (query) => {
 
 export const getServerSideProps = propsWithContainer(
   verifyToken(
-    async ({ query, container }) => {
-      const { id } = query;
+    async ({ query, container, authenticationToken }) => {
+      const id = authenticationToken.ward;
```
Comment (jivdhaliwal) [inline pages/wards/[id]/schedule-visit.js:257]: Should we call this wardId so it's clear what it relates to ?

Comment (antonyoneill) [reply pages/wards/[id]/schedule-visit.js:257]: We're actually going to remove the need for this in a next PR, so will kindly skip this

PR #145, opened 2020-04-29
Add new separate visitor name page

chubberlisk:
# What

Add a new separate visitor name page and a `BackLink` component.

# Why

We want to provide guidance as to what to expect and not do up front. Simply adding this to our current "Join a virtual visit" page will be overloading and may be missed. As a result, we want to split that page into first guidance and then the form to enter their name.

# Screenshots (Updated 14:07)

image

# Notes

Previous PR https://github.com/madetech/nhs-virtual-visit/pull/143, added the new start page for a visitor.

Next step is to update the link in the second text message to the visitor.

### DIFF
--- a/pages/visitors/[id]/name.js
+++ b/pages/visitors/[id]/name.js
@@ -1,0 +1,93 @@
+import React, { useCallback, useState } from "react";
+import { useRouter } from "next/router";
+import Button from "../../../src/components/Button";
+import FormGroup from "../../../src/components/FormGroup";
+import { GridRow, GridColumn } from "../../../src/components/Grid";
+import Hint from "../../../src/components/Hint";
+import Input from "../../../src/components/Input";
+import Layout from "../../../src/components/Layout";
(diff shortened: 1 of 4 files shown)


pages/visitors/[id]/name.js
```diff
@@ -0,0 +1,93 @@
...
+  });
+
+  return (
+    <Layout
+      title="What is your name? - Attend a virtual visit"
+      hasErrors={errors.length != 0}
+    >
+      <GridRow>
+        <GridColumn width="two-thirds">
+          <BackLink href={`/visitors/${router.query.id}/start`}>
```
Comment (steveknightmadetech) [inline pages/visitors/[id]/name.js:48]: Just a question after reading the position of backlinks here [(https://service-manual.nhs.uk/design-system/components/back-link)] it might be more consistent to put it at the bottom?

Comment (chubberlisk) [reply pages/visitors/[id]/name.js]: Yes, good spot!

We've updated `Layout` to take in a back link so we can easily change this in the future if necessary.

PR #151, opened 2020-04-29
Move cancel visit success

antonyoneill:
See #146


pages/wards/cancel-visit-success.js
```diff
@@ -0,0 +1,127 @@
...
+          callTime,
+          callDate,
+          callId,
+          error,
+          deleteError,
+        },
+      };
+    },
+    {
+      tokens: new TokenProvider(process.env.JWT_SIGNING_KEY),
```
chubberlisk, line 122: Maybe?

```suggestion
      tokens: container.getTokenProvider(),
```

antonyoneill: Not entirely sure it has access at this stage - Dan is going to be looking into the auth shortly so will probably make a sensible change here

PR #174, opened 2020-04-30
Added call password to joining a call

jbevan4:
# What
Added a call password to secure calls
# Why
Unsecured calls could of been hijacked
# Screenshots
N/A
# Notes
N/A

### DIFF
--- a/pages/visitors/[id]/start.js
+++ b/pages/visitors/[id]/start.js
@@ -57,15 +57,13 @@ const Start = ({ callId, error, queryPassword }) => {
 
 export const getServerSideProps = propsWithContainer(
   async ({ query, container, res }) => {
-    const { id, callPassword } = query;
-    const callId = id;
-    const queryPassword = callPassword ? callPassword : "";
+    const { id: callId, callPassword } = query;
 
     const verifyCallPassword = container.getVerifyCallPassword();
 
     const { validCallPassword, error } = await verifyCallPassword(
       callId,
-      queryPassword
+      callPassword
     );
 
     if (!validCallPassword) {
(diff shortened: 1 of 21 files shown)


pages/visitors/[id]/start.js
```diff
@@ -49,4 +54,17 @@ const Start = () => {
   );
 };
 
+export const getServerSideProps = propsWithContainer(
```
Comment (danielburnley) [inline pages/visitors/[id]/start.js:57]: Should probably add some tests for this :eyes:

pages/visitors/[id]/start.js
```diff
@@ -49,4 +54,17 @@ const Start = () => {
...
+  async ({ query, container }) => {
+    const { id, callPassword } = query;
+    const callId = id;
+    const queryPassword = callPassword ? callPassword : "";
+    const { scheduledCall, error } = await retrieveVisitByCallId(container)(
+      callId
+    );
+    const dbPassword = scheduledCall.callPassword;
+    return { props: { callId, error, queryPassword, dbPassword } };
+  }
```
Comment (antonyoneill) [inline pages/visitors/[id]/start.js:67]: Technically this would provide the end user with the real password if they looked in the page source

Comment (antonyoneill) [reply pages/visitors/[id]/start.js:67]: I would rather you redirect the user immediately if the password is incorrect. Ie do the check in this function (which runs server side) and if they fail then redirect them to `/error` similar to 
https://github.com/madetech/nhs-virtual-visit/blob/b9c428d21cf0be5cada288211c3315244dfc019a/pages/index.js#L6-L9

pages/visitors/[id]/start.js
Comment (jbevan4) [reply pages/visitors/[id]/start.js:57]: Added more tests :D

pages/visitors/[id]/start.js
Comment (jbevan4) [reply pages/visitors/[id]/start.js:73]: Copied this pattern 👍

PR #202, opened 2020-04-30
Add an admin flag to the API token if an admin code is used

jivdhaliwal:
# What
Add an admin flag to the API token if an admin code is used

# Why
Allows us to eventually support admin users

# Screenshots

# Notes
Requires an ADMIN_AUTH_CODES env variable


danielburnley:
We'll update the README in the follow up PR :tada:

PR #226, opened 2020-05-04
Refactored code and tests for displaying ward name

steveknightmadetech:
# What
Refactor of display ward names
# Why
To implement from Feedback from the last PR about the tests and implementation
# Screenshots

# Notes


pages/wards/visits.js
```diff
@@ -59,9 +59,10 @@ export const getServerSideProps = propsWithContainer(
       let { scheduledCalls, error } = await container.getRetrieveVisits()({
         wardId,
       });
-      let { ward, error: wardError } = await container.getWardById()(wardId);
+      let ward;
```
antonyoneill, line 62: This line was OK before hand :)

steveknightmadetech: Thought this way was cleaner :)

PR #228, opened 2020-05-05
Connect usecase to book visit

danielburnley:
# What
Update ward totals table when booking a visit

# Why

So that we can begin to understand the useage of the application

# Screenshots

# Notes

### DIFF
--- a/pageTests/api/book-a-visit.test.js
+++ b/pageTests/api/book-a-visit.test.js
@@ -13,6 +13,10 @@ describe("schedule-visit", () => {
     ward: "MEOW",
   }));
 
+  const updateWardVisitTotalsSpy = jest.fn(() => {
+    success: true;
+  });
+
   const getWardByIdSpy = jest.fn(() => ({
     ward: {
       id: 10,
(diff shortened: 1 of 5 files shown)


pageTests/api/book-a-visit.test.js
```diff
@@ -13,6 +13,10 @@ describe("schedule-visit", () => {
     ward: "MEOW",
   }));
 
+  const getUpdateWardVisitTotalsSpy = jest.fn(() => {
```
antonyoneill, line 16: This would be an `updateWardVisitTotalsSpy` right?

src/usecases/updateWardVisitTotals.js
```diff
@@ -3,7 +3,7 @@ export default ({ getDb }) => async ({ wardId, date }) => {
     const db = await getDb();
     const res = await db.any(
       "SELECT id, ward_id, total FROM ward_visit_totals WHERE ward_id = $1 AND total_date = $2",
-      [wardId, date.toISOString()]
+      [wardId, date]
     );
 
     const [existingTotal] = res;
```
antonyoneill, line 9: Do we need to be concerned about timing here? What if there are two ward staff adding visits at the same time? 

I guess at this stage this is an unlikely scenario

pageTests/api/book-a-visit.test.js
danielburnley: yes :eyes:

src/usecases/updateWardVisitTotals.js
danielburnley: (Summarising slack convo) For now this should be fine, longer term we'd probably opt for a more accurate performance system anyway so this will do for now

PR #235, opened 2020-05-05
Add eslint

antonyoneill:
# What
Add eslint & configure for react.

It will automatically run during commit, like prettier.

# Why
To give us quicker feedback on small mistakes, such as forgetting to import a module, misspelling a variable name, etc
# Screenshots

# Notes
Eslint and prettier may seem similar, but they are tools to achieve different jobs.
Prettier is more concerned with stylistic things such as semi-colons, single vs double quotes.
Eslint provides a more comprehensive error detection service, checking for misspellings, forgotten imports, variable name collisions, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,8 @@
   },
   "lint-staged": {
     "*.{js,css,md,yml}": [
-      "prettier --write"
+      "prettier --write",
+      "eslint --fix --ext .js"
     ]
   },
   "engines": {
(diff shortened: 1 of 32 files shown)


package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "format": "prettier --write '{pages,pageTests,src}/**/*.{js,css,md,yml}'",
     "format:check": "prettier --check '{pages,pageTests,src}/**/*.{js,css,md,yml}'",
+    "eslint": "eslint . --ext .js",
```
chubberlisk, line 9: Wondering if this command should be a generic name instead of the name of the linter? Thinking if we wanted to swap out ESLint for another we'd just change the command instead of both the command and script name.

```suggestion
    "lint": "eslint . --ext .js",
```

antonyoneill: YES :raised_hands: thank you

PR #248, opened 2020-05-11
Setup db-migrate

jivdhaliwal:
# What
Sets up [db-migrate](https://github.com/db-migrate/node-db-migrate) for use in the app. 

# Why
Allows us to manage and keep track of migrations on all environments

# Screenshots

# Notes
We are deploying with no migrations for now so that we can create the migrations table on all envs.
Once the migrations table is created, we will populate it with existing migrations so that when we run `dbmigrate up`, old migrations are not re-run on staging and production

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,9 +106,9 @@ Restart your PostgreSQL server and SSL will connections will be enabled
    ```bash
    createdb nhs-virtual-visit-dev
    ```
-2. Load the current schema into the database
+2. Run all migrations
    ```bash
-   cat db/schema.sql | psql nhs-virtual-visit-dev
+   npm run dbmigrate up
    ```
 3. Add the database URL as an environment variable in `.env`. On Linux you may need to provide a username and password.
    ```bash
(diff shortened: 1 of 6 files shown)


antonyoneill:
Nice! Would you mind putting something in the README about what it is and how to use it please?

README.md
```diff
@@ -162,6 +162,34 @@ Old scheduled calls are deleted after 24 hours. You can cleanup the table by run
 npm run cleandb
 ```
 
+## Applying database migrations
+
+Database migrations are managed with [db-migrate](https://github.com/db-migrate/node-db-migrate). To create a new migration
+
+```bash
+npm run dbmigrate create description-for-your-migration
```
chubberlisk, line 170: Is there a format we should follow or does `db-migrate` do this for us? Like the number at the beginning?

jivdhaliwal: There's no defined naming convention to follow but generally I would go with naming it in a way that describes the migration. For example add-column-name-to-table-name, create-table-name, drop-table-name, add-unique-constraint-to-table-name

PR #274, opened 2020-05-13
Added hospitals usecase

jbevan4:
# What
Added hospitals usecase
# Why
To further progress with adding hospitals
# Screenshots
N/A
# Notes
N/A

### DIFF
--- a/src/usecases/createHospital.js
+++ b/src/usecases/createHospital.js
@@ -1,0 +1,28 @@
+const createHospital = ({ getDb }) => async ({ name, trustId }) => {
+  const db = await getDb();
+
+  try {
+    console.log("Creating hospital for", name);
+    const createdHospital = await db.one(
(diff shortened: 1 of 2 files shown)


src/usecases/createHospital.js
```diff
@@ -0,0 +1,28 @@
+const createHospital = ({ getDb }) => async ({ name, trustId }) => {
+  const db = await getDb();
+
+  try {
+    console.log("Creating hospital for", name);
+    const createdHospital = await db.one(
+      `INSERT INTO hospitals
+          (id, name, trust_id)
```
chubberlisk, line 8: Do we need to specify the `id`?

jbevan4: We use this pattern in other places as we pass in ```default```. We could refactor to not use id, but I think that would be a separate PR to refactor it everywhere to be consistent :)

PR #278, opened 2020-05-13
Add hospital_id reference to wards

jivdhaliwal:
# What
Adds a hospital_id column to the wards table

# Why
Work towards allowing admins to add hospitals

# Screenshots

# Notes


StephenThomson:
LGTM. New line at the end of sql files would be my preference.

PR #281, opened 2020-05-14
Add trust name to to heading and page title of admin page

chubberlisk:
# What

Add trust name to to heading and page title of admin page.

# Why

We don't make it clear to the user what trust they are currently managing.

# Screenshots

image

# Notes

N/A

### DIFF
--- a/pages/admin.js
+++ b/pages/admin.js
@@ -52,12 +52,12 @@ export const getServerSideProps = propsWithContainer(
         authenticationToken.trustId
       );
 
-      const trustName = trustResponse.trust ? trustResponse.trust.name : null;
+      // const trustName = trustResponse.trust ? trustResponse.trust.name : null;
 
       return {
         props: {
           wards: wardsResponse.wards,
-          trust: { name: trustName },
+          trust: { name: trustResponse.trust?.name },
           wardError: wardsResponse.error,
           trustError: trustResponse.error,
         },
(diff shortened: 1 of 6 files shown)


pages/admin.js
```diff
@@ -36,12 +42,20 @@ const Admin = ({ error, wards }) => {
...
+      );
+      const trustResponse = await container.getRetrieveTrustById()(
         authenticationToken.trustId
       );
 
       return {
-        props: { wards: wards, error: error },
+        props: {
+          wards: wardsResponse.wards,
+          trust: trustResponse.trust,
```
chubberlisk, line 55: Offline chat: Be more explicit with the properties needed.

pages/admin.js
```diff
@@ -36,12 +45,22 @@ const Admin = ({ error, wards }) => {
...
+        authenticationToken.trustId
+      );
+
+      const trustName = trustResponse.trust ? trustResponse.trust.name : null;
 
       return {
-        props: { wards: wards, error: error },
+        props: {
+          wards: wardsResponse.wards,
+          trust: { name: trustName },
```
antonyoneill, line 60: I may be wrong, but I think this could be inlined as
```javascript
{ name: trustResponse.trust?.name }
```

I can't remember if the project is set up to use the newest syntax or not.

PR #284, opened 2020-05-14
Created UI for adding a hospital

jbevan4:
# What
Added the UI for adding a hospital
# Why
So admin users can add hospitals
# Screenshots
Screenshot 2020-05-14 at 15 22 29

# Notes
N/A


jbevan4:
There is an SQL query to sort that out @StephenThomson . Me and @jivdhaliwal have ran it on stage and prod so the hospitals table should be populated.

PR #292, opened 2020-05-18
Send message to a Slack channel when a milestone number of virtual visits

chubberlisk:
# What

Send a message to a Slack channel when a milestone number of virtual visits by:

- Adding a `Slack` gateway with a `postMessage` using Slack's web-api package
- Adding a `sendVisitsMilestoneNotification` usecase
- Updating `updateWardVisitTotals` usecase to call new usecase

# Why

To notify us when we've reached milestone and allow us to celebrate how far we've come from those first 48 hours.

# Screenshots

Screenshot 2020-05-18 at 08 59 27

# Notes

Next steps:

- Add a Slack bot to workspace
- Add environment variables to live


src/usecases/updateWardVisitTotals.js
```diff
@@ -9,10 +12,15 @@ export default ({ getDb }) => async ({ wardId, date }) => {
...
+      const newTotal = existingTotal.total + 1;
       await db.none("UPDATE ward_visit_totals SET total = $2 WHERE id = $1", [
         existingTotal.id,
-        existingTotal.total + 1,
+        newTotal,
       ]);
+
+      const sendVisitsMilestoneNotification = getSendVisitsMilestoneNotification();
+
+      await sendVisitsMilestoneNotification({ numberOfVisits: newTotal });
```
StephenThomson, line 23: I think we could do this asynchronously, I think I'd prefer not to make the app wait for this when it's non-essential. It may not matter as it's the last thing happening in this request but someone could add something later. I'd also like to make sure that a slack error won't cause the request to error.

src/usecases/sendVisitsMilestoneNotification.js
```diff
@@ -0,0 +1,18 @@
+const sendVisitsMilestoneNotification = ({ getSlack }) => async ({
+  numberOfVisits,
+}) => {
+  if (numberOfVisits % 500 === 0) {
```
StephenThomson, line 4: Maybe the 500 should be an env variable or at least a const at the top of file. I hope that the app gets a lot of use then 500 will be too often!

chubberlisk: Ah good shout, I like this idea!

src/usecases/updateWardVisitTotals.js
chubberlisk: Ah yush, this makes sense! An error in Slack shouldn't affect this usecase as it will be caught in the gateway which then just returns an error key to the usecase. Any improvement ideas or is this okay?

antonyoneill: Is this the right total? Or is this intentionally looking for 500 visits in a single ward?

chubberlisk: Ah, ah, ah... 🤦 You're right, this is for a single ward. Good spot! 🕵️‍♂️

Looks like I need to think about this more in my own time.

chubberlisk:
Closing the PR so it's not noisy amongst other PRs. Need to figure how else to do this after https://github.com/madetech/nhs-virtual-visit/pull/292#discussion_r426734395. 🤔

PR #294, opened 2020-05-18
Swap hospital name with id

jbevan4:
# What
Swapped the use of hospital name with hospital id
# Why
So we can use relationship not stored text value on wards
# Screenshots
NA
# Notes
NA

### DIFF
--- a/src/usecases/retrieveWardById.js
+++ b/src/usecases/retrieveWardById.js
@@ -9,10 +9,11 @@ const retrieveWardById = ({ getDb }) => async (wardId, trustId) => {
         (
           SELECT
             name
-          from
+          FROM
             hospitals
-          where
+          WHERE
             id = wards.hospital_id
+            AND trust_id = $2
         ) as hospital_name,
         wards.hospital_id as hospital_id
       FROM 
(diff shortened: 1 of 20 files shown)


src/usecases/retrieveWardById.js
```diff
@@ -3,14 +3,15 @@ const retrieveWardById = ({ getDb }) => async (wardId, trustId) => {
   console.log("Retrieving ward for  ", wardId);
   try {
     const ward = await db.oneOrNone(
-      "SELECT * FROM wards WHERE id = $1 AND trust_id = $2 LIMIT 1",
+      "SELECT wards.id as ward_id, wards.name as ward_name, (SELECT name from hospitals where id=wards.hospital_id) as hospital_name, wards.hospital_id as hospital_id FROM wards WHERE wards.id = $1 AND wards.trust_id = $2 LIMIT 1",
```
antonyoneill, line 6: Can you add some whitespace to this please, will be easier to read and spot the subquery!

src/usecases/retrieveWardById.js
```diff
@@ -3,14 +3,35 @@ const retrieveWardById = ({ getDb }) => async (wardId, trustId) => {
...
+      `SELECT
+        wards.id as ward_id,
+        wards.name as ward_name,
+        (
+          SELECT
+            name
+          from
+            hospitals
+          where
+            id = wards.hospital_id
```
jivdhaliwal, line 15: I believe we need to update this query to include the trust ID when retrieving the hospital (same for https://github.com/madetech/nhs-virtual-visit/pull/294/files#diff-d8c3299b5e0b2cd1d5c6fc7d55081ad5R9-R15)

jbevan4: I've pushed what I think you mean :)

PR #295, opened 2020-05-18
Split ward visits panel list into today and upcoming

chubberlisk:
# What

Split ward visits panel list into today and upcoming using new helper functions.

# Why

To make it the visits clearer to the ward staff what's important now.

# Screenshots

## Before

image

## After

image

# Notes

Next step is to add the beside which show how long until the visit.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11395,6 +11395,11 @@
         "minimist": "^1.2.5"
       }
     },
+    "mockdate": {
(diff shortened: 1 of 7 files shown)


StephenThomson:
Logged in as 'TEST1' on the review app and tried to add a visit and the book button is failing. This is something I have seen in my branch too but thought it was an error I'd introduced.

PR #301, opened 2020-05-19
Add visit from now time to accordionised visits

chubberlisk:
# What

Add visit from now time to accordionised visits.

# Why

To provide ward staff an indication of how long a visit is in.

# Screenshots

## Before

image

## After

image

# Notes

Would recommend using the review app to test:

1. Book a visit in a few minutes time
2. Go to the visits page with `?showAccordion=1`
3. You should notice the time decrease in a minute

### DIFF
--- a/src/components/VisitsPanelList/index.js
+++ b/src/components/VisitsPanelList/index.js
@@ -116,6 +116,8 @@ const VisitPanelList = ({ visits, title }) => {
         </ul>
       </div>
     );
+  } else {
+    return null;
   }
 };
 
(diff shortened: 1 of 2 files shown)


src/components/VisitsPanelList/index.js
```diff
@@ -1,11 +1,34 @@
...
 import "./styles.scss";
 import Router from "next/router";
 import formatDateAndTime from "../../helpers/formatDateAndTime";
 import VisitSummaryList from "../VisitSummaryList";
 import { GridRow, GridColumn } from "../Grid";
 import filterTodaysVisits from "../../helpers/filterTodaysVisits";
 import filterUpcomingVisits from "../../helpers/filterUpcomingVisits";
+import moment from "moment";
+
+const Beside = ({ callDateAndTime }) => {
```
jivdhaliwal, line 11: Should we extract this out to its own component file ?

chubberlisk: That's a good shout! I like dis. Me do dis.

chubberlisk: Created a `TimeFromNow` component which will return the time from now that will update every 1 minute by default but can be set.

PR #303, opened 2020-05-19
Database testing

jivdhaliwal:
# What
Implement testing that hits the database

# Why
So that we can create full e2e tests

# Screenshots

# Notes

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,7 +4,17 @@ orbs:
   node: circleci/node@1.1.6
 
 jobs:
-  build-and-test:
+  build:
+    executor:
+      name: node/default
+    steps:
+      - checkout
+      - node/with-cache:
+          steps:
+            - run: npm install
+            - run: npm rebuild node-sass
+            - run: npm run build
+  test:
     executor:
       name: node/default
     steps:
@@ -14,9 +24,9 @@ jobs:
             - run: npm install
             - run: npm rebuild node-sass
             - run: npm run test:ci
-            - run: npm run build
 
 workflows:
   build-and-test:
     jobs:
-      - build-and-test
+      - build
+      - test
--- a/circle.yml
+++ b/circle.yml
@@ -15,6 +15,17 @@ jobs:
             - run: npm rebuild node-sass
             - run: npm run build
   test:
+    docker:
+      - image: circleci/node:12
+      - environment:
+          NODE_ENV: test
+          POSTGRES_USER: root
+          POSTGRES_PASSWORD:
+          POSTGRES_DB: nhs-virtual-visit-test
+        image: "circleci/postgres:12.2"
+    environment:
+      NODE_ENV: test
+      TEST_DATABASE_URL: postgresql://root@localhost/nhs-virtual-visit-test
     executor:
       name: node/default
     steps:
@@ -23,7 +34,8 @@ jobs:
           steps:
             - run: npm install
             - run: npm rebuild node-sass
-            - run: npm run test:ci
+      - run: npm run dbmigratetest up
+      - run: npm run test:ci
 
 workflows:
   build-and-test:
--- a/src/testUtils/truncateAllTables.js
+++ b/src/testUtils/truncateAllTables.js
@@ -1,0 +1,8 @@
+export default async ({ getDb }) => {
+  const db = await getDb();
+  await db.any("DELETE from ward_visit_totals");
+  await db.any("DELETE from scheduled_calls_table");
+  await db.any("DELETE from wards");
+  await db.any("DELETE from hospitals");
+  await db.any("DELETE from trusts");
+};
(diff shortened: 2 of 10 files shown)


src/usecases/createWard.test.js
```diff
@@ -30,4 +36,42 @@ describe("createWard", () => {
...
+        hospitalName: "Test Hospital",
+        code: "WardCode",
+        trustId: trust.id,
+        hospitalId: hospital.id,
+      };
+
+      const { wardId, error } = await createWard(container)(request);
+
+      expect(wardId).toBeTruthy;
+      expect(error).toBeNull();
```
StephenThomson, line 74: It might be nice to actually verify that the ward is present in the db rather than trusting that the createWard honestly reports its success.

src/testUtils/truncateAllTables.js
```diff
@@ -0,0 +1,8 @@
+export default async ({ getDb }) => {
+  const db = await getDb();
+  await db.any("DELETE from scheduled_calls_table");
+  await db.any("DELETE from wards");
```
antonyoneill, line 4: I tried deleting from wards before, if a ward has a visit counted against it I think this will fail because of foreign keys - just something to bare in mind

jivdhaliwal: Calls are deleted first to avoid the constraint errors (and wards are deleted before hospitals etc)

antonyoneill: My issue was I had calls counted in `ward_visit_totals` against wards so couldn't delete a ward until I'd deleted the row from the totals table

jivdhaliwal: Ah I didn't see the constraint on ward_visit_totals. That needs to go first. Or we just delete while ignoring constraints

jivdhaliwal:
> * You mentioned that the cleanup takes time and slows down the tests - I'm curious how much it does slow them down by? Of course we should balance testing against a real db with speed of feedback.
>   I'm cautious that having a requirement to run these database tests will slow down the general unit tests of the application.
Given the size of our app and test suite, I believe the speed difference will be negligible

>   I'm thinking that with the introduction of these tests we should refactor the db interaction out into the `DatabaseGateway` (like we do for notify) and have these tests only run when the contract tests are run.
Are you suggesting another gateway? (like TestDatabaseSetupGateway)
And should these tests be moved to contract tests or moved under a new integration config?

> * The underlying database technology becomes less of a concern by using the existing usecases to insert fixtures. This also has other benefits:
>
>   * The fixtures can be used with a mocked out database (faster execution)
>   * If the data structure changes in the future, tables renamed, etc this is one less place to change

Agreed. Taken out the hospital function and replaced with the usecase. I'd like to leave the trust one in there for now until we have a usecase for creating Trusts (soon).

antonyoneill:
> Given the size of our app and test suite, I believe the speed difference will be negligible

perfect :ok_hand: 

> Are you suggesting another gateway? (like TestDatabaseSetupGateway)
> And should these tests be moved to contract tests or moved under a new integration config?

No - I'm thinking that there should be no db interactions, or SQL statements within `src/usecases/`. This could be achieved by refactoring any interactions into the existing `DatabaseGateway`. i.e. in `createAWard` you would end up with something like

```javascript
const db = await getDb();

const result = await db.createWard({wardName, hospitalId, ...});

if (result.success) {
...
} 
```

Then the only test that uses the database is a DatabaseGateway.contractTest.js (like GovNotify - but probably wouldn't call it contractTest, rather externalTest or something, idk)

> Agreed. Taken out the hospital function and replaced with the usecase. I'd like to leave the trust one in there for now until we have a usecase for creating Trusts (soon).

:ok_hand:

circle.yml
```diff
@@ -13,10 +13,32 @@ jobs:
...
             - run: npm run build
+  test:
+    docker:
+      - image: circleci/node:12
+      - environment:
+          NODE_ENV: test
+          POSTGRES_USER: root
+          POSTGRES_PASSWORD:
+          POSTGRES_DB: nhs-virtual-visit-test
+        image: "circleci/postgres:12.2"
```
jbevan4, line 25: Can you move this into a ```- image:``` similar to ```- image: circleci/node:12```

jbevan4: So maybe the structure could be 
```- image: circleci/node:12
     - image: circleci/postgres:12.2
```

jivdhaliwal: Will take a crack at it. The docs I were following had the image indented which I assumed was doing some linking under the hood.

jbevan4: Cool no worries :)

jivdhaliwal: I just tested moving it around and it causes a build error. The funky indentation is intentional 🤷

antonyoneill: Yaml has some very strange, I think ambiguous, syntax around objects :(

PR #305, opened 2020-05-19
Use app container for token provider

StephenThomson:
# What

Previously a custom container was used to pass in token provider, now we use the app container.

# Why

Consistency
To support the work to ensure login is not possible with tokens from archived wards.

# Screenshots

# Notes

### DIFF
--- a/pageTests/wards/visit-start.test.js
+++ b/pageTests/wards/visit-start.test.js
@@ -16,7 +16,17 @@ describe("visit-start", () => {
 
   describe("getServerSideProps", () => {
     it("redirects to login page if not authenticated", async () => {
-      await getServerSideProps({ req: anonymousReq, res });
+      const authenticationToken = {
+        foo: true,
+      };
+      const tokenProvider = {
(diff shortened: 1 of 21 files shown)


pageTests/wards/visit-start.test.js
```diff
@@ -16,7 +16,17 @@ describe("visit-start", () => {
...
+      const authenticationToken = {
+        foo: true,
+      };
+      const tokenProvider = {
+        validate: jest.fn(() => authenticationToken),
+      };
+      const container = {
+        getTokenProvider: () => tokenProvider,
+      };
+      console.log(container);
```
lukemorton, line 28: I spy the debug

StephenThomson: 😞

StephenThomson:
I think there might be a problem, need to have a look with someone with heroku access tomorrow

StephenThomson:
The problem was just that the Notify API key used in staging will only text team members and I wasn't a team member.

PR #313, opened 2020-05-20
Cypress e2e tests

jivdhaliwal:
# What
Adds support for e2e testing using cypress

# Why
Allows us to test the whole app from start to finish

# Screenshots

# Notes


src/gateways/Database/index.js
```diff
@@ -14,7 +14,7 @@ export default (() => {
           options.connectionString = process.env.TEST_DATABASE_URL;
         }
 
-        if (process.env.NODE_ENV !== "test") {
+        if (process.env.NODE_ENV === "production") {
```
antonyoneill, line 17: Just checking this is the correct way around?

If it's production we almost certainly want to check the SSL cert? I may have missed something

circle.yml
```diff
@@ -42,3 +59,34 @@ workflows:
...
+          # installed by the "install" job
+          requires:
+            - Linux install
+          record: false
+          parallel: false # load balance all tests
+          parallelism: 1 # across 3 machines
+          group: "1x-electron on CircleCI" # name this group
+          post-install:
+            - run: npm run dbmigratetest up
+          start: "npm run dev" # start server before running tests
```
antonyoneill, line 91: Maybe an `npm run start` would have been a better state to test against?

src/gateways/Database/index.js
jivdhaliwal: Ye this change means that if the env is production, require SSL.

If we're only requiring SSL in production then I guess we can remove the `{ rejectUnauthorized: false }` and just replace it right `true` ? (Was this added to make it work with dev signed certs ?)

circle.yml
jivdhaliwal: We would need to get the database working with SSL on circleCI before we can make that change (previous attempts didn't go so well)

PR #320, opened 2020-05-21
Refactor token user type

jivdhaliwal:
# What
Adds a new 'type' key to the token to distinguish user types. 

# Why
We currently use an 'admin' boolean to distinguish users from admins. This is not open to extending to more types in the future.

# Screenshots

# Notes

### DIFF
--- a/src/usecases/adminIsAuthenticated.js
+++ b/src/usecases/adminIsAuthenticated.js
@@ -8,7 +8,7 @@ export default ({ getTokenProvider }) => (requestCookie) => {
 
     const validatedToken = tokenProvider.validate(token);
 
-    if (token && !validatedToken.admin) {
+    if (token && validatedToken.type !== "trustAdmin") {
       return false;
     }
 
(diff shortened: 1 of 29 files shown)


src/usecases/adminIsAuthenticated.js
```diff
@@ -8,7 +8,7 @@ export default ({ getTokenProvider }) => (requestCookie) => {
 
     const validatedToken = tokenProvider.validate(token);
 
-    if (token && !validatedToken.admin) {
+    if (token && validatedToken.type !== "trustAdmin") {
```
StephenThomson, line 11: We could define the string that corresponds to a type in a single place and then reference it when needed.

jivdhaliwal: Ye good point it would be nice to store them as constants somewhere

jivdhaliwal: Moved the token types out to a constant in 62584eee351103992

PR #322, opened 2020-05-21
Default re-booking date to day following original call

neilkidd:
# What
When re-booking a call the day for the next call is defaulted to the day after the original call.

# Why
A UX improvement for ease of use.

# Screenshots
N/A

# Notes
N/A


pageTests/wards/book-a-visit.test.js
```diff
@@ -1,5 +1,10 @@
 import { getServerSideProps } from "../../pages/wards/book-a-visit";
 
+// TODO: This needs to be moved once the verifyToken logic is in the container..
```
jivdhaliwal, line 3: This can be removed since we're using the tokenProvider in the tests

StephenThomson: Done

PR #324, opened 2020-05-21
Update filtering of visits

chubberlisk:
# What

Update filtering of visits so that:

- Today is 1 hour ago until the end of the day
- Past is between 1 hour ago and 12 hours

# Why

This makes it consistent to what we have on the visits table and is the expected behaviour.

# Screenshots

N/A

# Notes

~Currently the user will have to refresh the page to see it a visit more than an hour ago "move" the past.~ 😢

### DIFF
--- a/src/helpers/filterPastVisits.test.js
+++ b/src/helpers/filterPastVisits.test.js
@@ -3,25 +3,31 @@ import filterPastVisits from "./filterPastVisits";
 
 describe("filterPastVisits", () => {
   beforeEach(() => {
-    MockDate.set(new Date("2020-05-18"));
+    MockDate.set(new Date("2020-05-18 13:00"));
   });
 
   afterEach(() => {
     MockDate.reset();
   });
 
-  it("returns visits that were scheduled in the past", () => {
+  it("returns visits that were scheduled between 1 hour ago and 12 hours ago", () => {
     const visits = [
-      { callTime: "2020-05-17 13:00:00" },
-      { callTime: "2020-05-17 23:59:59" },
-      { callTime: "2020-05-18 00:00:00" },
-      { callTime: "2020-05-18 00:00:01" },
+      { callTime: "2020-05-17 15:00:00" },
+      { callTime: "2020-05-18 00:59:59" },
+      { callTime: "2020-05-18 01:00:00" },
+      { callTime: "2020-05-18 09:00:00" },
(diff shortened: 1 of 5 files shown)


src/helpers/filterTodaysVisits.test.js
```diff
@@ -3,24 +3,27 @@ import filterTodaysVisits from "./filterTodaysVisits";
...
     MockDate.reset();
   });
 
-  it("returns visits that are scheduled for today", () => {
+  it("returns visits that are scheduled between 1 hour ago until end of the day", () => {
     const visits = [
       { callTime: "2020-05-17 23:59:59" },
+      { callTime: "2020-05-18 09:00:00" },
+      { callTime: "2020-05-18 11:59:59" },
+      { callTime: "2020-05-18 12:00:00" },
```
jivdhaliwal, line 18: Is it worth including a visit at 12:00:01 to show what visits around this point are included ? (we have 1 second before but not 1 second after)

src/helpers/filterPastVisits.test.js
```diff
@@ -3,25 +3,30 @@ import filterPastVisits from "./filterPastVisits";
...
-      { callTime: "2020-05-17 13:00:00" },
-      { callTime: "2020-05-17 23:59:59" },
-      { callTime: "2020-05-18 00:00:00" },
-      { callTime: "2020-05-18 00:00:01" },
+      { callTime: "2020-05-17 15:00:00" },
+      { callTime: "2020-05-18 00:59:59" },
+      { callTime: "2020-05-18 01:00:00" },
+      { callTime: "2020-05-18 09:00:00" },
+      { callTime: "2020-05-18 11:59:59" },
+      { callTime: "2020-05-18 12:00:00" },
```
jivdhaliwal, line 20: Is it worth including a visit at 12:00:01 to show what visits around this point are included ? (we have 1 second before but not 1 second after)

src/helpers/filterTodaysVisits.test.js
chubberlisk: Makes sense, updated the PR!

src/helpers/filterPastVisits.test.js
chubberlisk: Makes sense, updated the PR!

PR #329, opened 2020-05-22
Add support for admin users

jivdhaliwal:
# What
Adds support for logging in as an Admin. Admins will have the ability to create Trusts

# Why
This will put more control in the hands of the end user so that they can manage the app

# Screenshots

# Notes


jivdhaliwal:
> Looks great!
> 
> Optional refactor suggestion: Now that we have 3 verify[Usertype]IsAuthenticated methods I think we could think about refactoring. They already return a type so a single method that handles all 3 cases shouldn't create too much change in other places.
> 
> unicorngif

Ye agreed the duplication has got to a point where it makes sense to refactor

PR #332, opened 2020-05-22
Added the usecase for adding a trust

jbevan4:
# What
Added a trust usecase
# Why
So admins can add trusts
# Screenshots
NA
# Notes
NA

### DIFF
--- a/src/usecases/createTrust.contractTest.js
+++ b/src/usecases/createTrust.contractTest.js
@@ -1,0 +1,31 @@
+import createTrust from "./createTrust";
+import AppContainer from "../containers/AppContainer";
+import truncateAllTables from "../testUtils/truncateAllTables";
+
+describe("createTrust contract tests", () => {
+  const container = AppContainer.getInstance();
+
+  beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


src/usecases/createTrust.contractTest.js
```diff
@@ -0,0 +1,31 @@
...
+
+  it("creates a trust in the db when valid", async () => {
+    const request = {
+      name: "Defoe Trust",
+      adminCode: "adminCode",
+    };
+
+    const { trustId, error } = await createTrust(container)(request);
+
+    expect(trustId).not.toBeNull();
```
jivdhaliwal, line 27: What do you think about updating this expectation so that it checks that the created trust matches the params we passed in ? In the create ward contract test [we retrieve the created ward](https://github.com/madetech/nhs-virtual-visit/blob/88833924125d70c94104c7087edf1609a9716a0f/src/usecases/createWard.contractTest.js#L41-L48) and check it is correct

jbevan4: I missed the usecase for trustId, will do that now.

PR #339, opened 2020-05-26
Implement email validation

StephenThomson:
# What
Implement email validation helper
# Why
To support adding email notification of scheduled visits
# Screenshots

# Notes


src/helpers/validateEmailAddress.test.js
```diff
@@ -1,8 +1,11 @@
...
 
 describe("validateEmailAddress", () => {
-  it("accepts an email address", () => {
-    const actual = validateEmailAddress("hello@goodbye.com");
-    expect(actual).toEqual(true);
+  it("accepts a valid email address", () => {
+    expect(validateEmailAddress("hello@goodbye.com")).toEqual(true);
+  });
+
+  it("reject and invalid email address", () => {
```
chubberlisk, line 8:
```suggestion
  it("reject an invalid email address", () => {
```

src/helpers/validateEmailAddress.test.js
```diff
@@ -1,8 +1,11 @@
 import validateEmailAddress from "./validateEmailAddress";
 
 describe("validateEmailAddress", () => {
-  it("accepts an email address", () => {
-    const actual = validateEmailAddress("hello@goodbye.com");
-    expect(actual).toEqual(true);
+  it("accepts a valid email address", () => {
+    expect(validateEmailAddress("hello@goodbye.com")).toEqual(true);
```
chubberlisk, line 5: Wondering if it's worth adding more email examples here? 🤔

src/helpers/validateEmailAddress.test.js
```diff
@@ -1,8 +1,63 @@
...
+  "japanese-info@例え.テスト",
+];
+
+const invalidEmailAddresses = [
+  "email@[123.123.123.123]",
+  "plainaddress",
+  "@no-local-part.com",
+  "Outlook Contact <outlook-contact@domain.com>",
+  "no-at.domain.com",
+  "no-tld@domain",
```
antonyoneill, line 28: ^ This is technically a valid email address..

src/helpers/validateEmailAddress.test.js
```diff
@@ -1,8 +1,63 @@
 import validateEmailAddress from "./validateEmailAddress";
 
+const validEmailAddresses = [
+  "email@domain.com",
+  "email@domain.COM",
+  "firstname.lastname@domain.com",
+  "firstname.o'lastname@domain.com",
```
antonyoneill, line 7: ^ thank you for supporting my cause here

src/helpers/validateEmailAddress.js
```diff
@@ -1,4 +1,17 @@
...
+  "a-zA-Z0-9àèìòùÀÈÌÒÙáéíóúýÁÉÍÓÚÝâêîôûÂÊÎÔÛãñõÃÑÕäëïöüÿÄËÏÖÜŸçÇßØøÅåÆæœ" + // Number and letters including accented characters
+  "\u3000-\u303F\u3040-\u309F\u30A0-\u30FF\uFF00-\uFFEF\u4E00-\u9FAF\u2605-\u2606\u2190-\u2195\u203B"; // Chinese/Japanese/Korean characters
+
+const EMAIL_VALIDATION_STRING =
+  `^[${ALPHANUMERIC}.!#$%&'*+/=?^_\`{|}~-]` + // Local part
+  `+@[${ALPHANUMERIC}](?:[${ALPHANUMERIC}-]` + // Domain name
+  `{0,61}[${ALPHANUMERIC}])?(?:\\.` + // Allow periods in domain name
+  `[${ALPHANUMERIC}](?:[${ALPHANUMERIC}-]{0,61}[${ALPHANUMERIC}])?)*\\.` +
+  `[${ALPHANUMERIC}](?:[${ALPHANUMERIC}-]{0,61}` + // End of domain
+  `[${ALPHANUMERIC}])$`;
```
antonyoneill, line 11: This is pretty gross.. Is there an email address validation library we can use instead? :/

package.json
```diff
@@ -32,6 +32,7 @@
     "db-migrate": "^0.11.11",
     "db-migrate-pg": "^1.2.1",
     "dotenv-load": "^2.0.0",
+    "email-validator": "^2.0.4",
```
jivdhaliwal, line 35: Do we need this package any more given the validateEmailAddress helper ?

src/helpers/validateEmailAddress.js
antonyoneill: I've just seen you have installed one in the package.json - did it not work out?

neilkidd:
LGTM :+1:

src/helpers/validateEmailAddress.test.js
antonyoneill: (no action required)

package.json
jbevan4, line 35: Does this library not validate for us then? If so why do we need it?

chubberlisk: We no longer need it, forgot to remove! 😅

src/helpers/validateEmailAddress.js
chubberlisk: Discussed on Slack 💬

antonyoneill: For propsperity

> 
> Steve Thomson  14 minutes ago
> The reason we have done it manually is that there were problems with GovNotify validation not being consistent with validation libraries on the DFE teacher training project. This would cause bad user experience if the form said valid and then we got a rejection from GovNotify. As we already had an implementation we could steal we went for it. https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/960/files
> :this:
> 1
> 
> 
> Antony O'Neill  11 minutes ago
> :thinking_face:  Do we use error messages from GovNotify to determine validity of phone numbers?
> 
> Steve Thomson  10 minutes ago
> Yes and no. We validate using a library but then GN  will throw an error later if it doesn't like the number. You can see this happening if you try a non-team phone number on staging. (edited) 
> 
> Steve Thomson  10 minutes ago
> Which would return a 400 from the api to the client, which we then don't handle super gracefully imo. I think I made a ticket to look at it.
> :+1:
> 1
> 
>

PR #342, opened 2020-05-26
Added tokens table

jbevan4:
# What
Add tokens table
# Why
So we can store refresh tokens
# Screenshots
NA
# Notes
NA

### DIFF
--- a/db/migrations/sqls/20200526142326-add-refresh-tokens-up.sql
+++ b/db/migrations/sqls/20200526142326-add-refresh-tokens-up.sql
@@ -1,0 +1,4 @@
+CREATE TABLE tokens (
+  id serial PRIMARY KEY,
+  token_value varchar(255) NOT NULL UNIQUE
+);
--- a/db/migrations/sqls/20200526142326-add-refresh-tokens-up.sql
+++ b/db/migrations/sqls/20200526142326-add-refresh-tokens-up.sql
@@ -1,4 +1,4 @@
-CREATE TABLE tokens (
+CREATE TABLE refresh_tokens (
   id serial PRIMARY KEY,
-  token_value varchar(255) NOT NULL UNIQUE
+  value varchar(255) NOT NULL UNIQUE
 );
--- a/db/migrations/20200526142326-add-refresh-tokens.js
+++ b/db/migrations/20200526142326-add-refresh-tokens.js
@@ -1,0 +1,59 @@
+"use strict";
+
+var dbm;
+var type;
+var seed;
+var fs = require("fs");
+var path = require("path");
+var Promise;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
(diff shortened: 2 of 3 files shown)


db/migrations/sqls/20200526142326-add-refresh-tokens-up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE tokens (
+  id serial PRIMARY KEY,
+  token_value varchar(255) NOT NULL UNIQUE
```
jivdhaliwal, line 3: Since this is the 'tokens' table, could this be renamed to simply 'value' ?

db/migrations/sqls/20200526142326-add-refresh-tokens-up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE tokens (
```
jivdhaliwal, line 1: Do we want to be explicit here and call this refresh_tokens or should we stick with the more generic tokens which doesn't assume any specific implementation ?

db/migrations/sqls/20200526142326-add-refresh-tokens-up.sql
neilkidd: +1

jbevan4: I tried `value` but I was getting an error from postgres. Will try it again later.

db/migrations/sqls/20200526142326-add-refresh-tokens-up.sql
jbevan4: refresh_tokens make more sense, changing it.

db/migrations/sqls/20200526142326-add-refresh-tokens-up.sql
jivdhaliwal: Ah value is a reserved keyword (https://www.postgresql.org/docs/8.1/sql-keywords-appendix.html)

jbevan4: it's worked when I've changed it...

db/migrations/sqls/20200526142326-add-refresh-tokens-up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE refresh_tokens (
+  id serial PRIMARY KEY,
+  value varchar(255) NOT NULL UNIQUE
```
jivdhaliwal, line 3: Assuming we are going to be searching this table by the token, do we need to index the value column? 

And also is a 255 length long enough for the tokens we will be generating ?

jbevan4: 255 was long enough in testing. Depending on how much we change the token though this will increase/decrease. 

We should probably add an idea. Will add another todo.

jbevan4: Because this is the refreshToken, it should be smaller than our actual token, as it only needs a subset of fields.

PR #343, opened 2020-05-26
Added add refresh tokens usecase

jbevan4:
# What
Added usecase for adding refresh tokens
# Why
So we can add refresh tokens
# Screenshots
NA
# Notes
NA

### DIFF
--- a/src/usecases/createRefreshToken.js
+++ b/src/usecases/createRefreshToken.js
@@ -1,0 +1,27 @@
+const createRefreshToken = ({ getDb }) => async ({ refreshToken }) => {
+  const db = await getDb();
+
+  try {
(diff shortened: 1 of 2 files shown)


jbevan4:
Yeah I couldn't think of a better name. I just adopted the pattern from other files in our repo. I can rename if that's the case. 

The token is generated by the server and needs to be persisted. "Save" feels like the right word but unsure.

StephenThomson:
Yeah that's true... I guess usually we know we aren't "creating" a visit anywhere as it's just a record whereas with a token the creation will happen in code. IDK what the right answer is 🤷

jbevan4:
Going with the quicker win, closing this PR! :D

PR #344, opened 2020-05-26
Govnotify integration email send usecase

neilkidd:
# What
Implements the sendEmail useCase

# Why
In readiness for wiring in

# Screenshots

# Notes
Not wired in yet. No functional change.
Needs contract tests to be considered

### DIFF
--- a/__mocks__/notifications-node-client.js
+++ b/__mocks__/notifications-node-client.js
@@ -69,4 +69,6 @@ export class NotifyClient {
       id: "test-sms-return-response-id",
     });
   });
+
+  sendEmail = jest.fn(() => {});
 }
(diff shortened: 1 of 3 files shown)


__mocks__/notifications-node-client.js
```diff
@@ -69,4 +69,17 @@ export class NotifyClient {
       id: "test-sms-return-response-id",
     });
   });
+
+  sendEmail = jest.fn((templateId) => {
```
jivdhaliwal, line 73: Should we also write a contract test to make sure this behaves as we expect it to when we hit the actual library ?

neilkidd: Yes - I'll add that in to the next PR. (Contract tests)

PR #345, opened 2020-05-26
Refactor test teardown

jivdhaliwal:
# What
Extracts the database closing test teardown step to a separate file

# Why
Ensures that the database connection is closed correctly after all tests have run without having to include it in every test

# Screenshots

# Notes
This change also runs contract tests on a single process when you run test:contract, as the database clearing causes issues when run in parallel (due to the fact that we only have 1 test database)

### DIFF
--- a/jest.contract.config.js
+++ b/jest.contract.config.js
@@ -126,7 +126,7 @@ module.exports = {
   // setupFiles: [],
 
   // A list of paths to modules that run some code to configure or set up the testing framework before each test
-  setupFilesAfterEnv: ["<rootDir>/testSetup.js"],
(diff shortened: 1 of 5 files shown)


antonyoneill:
What if the developer ran a single contract test via CLI that didn't use the database? Would this be safe or blow up? I can't tell - worth a go

jivdhaliwal:
> What if the developer ran a single contract test via CLI that didn't use the database? Would this be safe or blow up? I can't tell - worth a go

Running a single test that doesn't use the database will work normally (it doesn't complain if there are no connections to terminate)

PR #352, opened 2020-05-27
Refresh token expiry

jbevan4:
# What
Refreshed the "admin" token if within a refresh timeframe
# Why
So users are not logged out if they are using the app
# Screenshots
NA
# Notes
NA

### DIFF
--- a/src/usecases/refreshToken.js
+++ b/src/usecases/refreshToken.js
@@ -1,0 +1,24 @@
+import moment from "moment";
+
+export default ({ getTokenProvider }) => (token) => {
+  const expiresAt = new Date(token.exp * 1000);
+  if (
+    moment(moment()).isBetween(
+      moment(expiresAt).subtract(3, "hours"),
+      expiresAt
+    )
+  ) {
+    const tokenProvider = getTokenProvider();
+    const refreshedEncodedToken = tokenProvider.generate({
+      wardId: token.wardId,
+      wardCode: token.wardCode,
+      trustId: token.trustId,
+      type: token.type,
+    });
+
+    const refreshedToken = tokenProvider.validate(refreshedEncodedToken);
+    return { refreshedToken, refreshedEncodedToken, isTokenRefreshed: true };
+  } else {
+    return { isTokenRefreshed: false };
+  }
+};
--- a/src/usecases/refreshToken.test.js
+++ b/src/usecases/refreshToken.test.js
@@ -1,0 +1,52 @@
+import refreshToken from "./refreshToken";
+import moment from "moment";
+
+describe("verifyAdminToken", () => {
+  it("it doesn't refresh the token if outside the refresh timeframe", () => {
+    const authenticationToken = {
+      type: "admin",
+      exp: moment().subtract(4, "hours").unix(),
+    };
+
+    const tokenProvider = {
+      validate: jest.fn(() => authenticationToken),
+      generate: jest.fn(() => "encodedToken"),
+    };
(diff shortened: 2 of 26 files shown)


src/usecases/refreshToken.js
```diff
@@ -0,0 +1,24 @@
+import moment from "moment";
+
+export default ({ getTokenProvider }) => (token) => {
+  const expiresAt = new Date(token.exp * 1000);
```
chubberlisk, line 4: Sorry, probs a silly question: What does the `token.exp * 1000` mean?

src/usecases/refreshToken.js
```diff
@@ -0,0 +1,24 @@
...
+  if (
+    moment(moment()).isBetween(
```
chubberlisk, line 6: Do we need to wrap `moment` in `moment`? 🤔

src/usecases/refreshToken.test.js
```diff
@@ -0,0 +1,52 @@
+import refreshToken from "./refreshToken";
+import moment from "moment";
+
+describe("verifyAdminToken", () => {
```
chubberlisk, line 4:
```suggestion
describe("refreshToken", () => {
```

src/usecases/refreshToken.test.js
```diff
@@ -0,0 +1,52 @@
...
+  it("it doesn't refresh the token if outside the refresh timeframe", () => {
```
chubberlisk, line 5:
```suggestion
  it("doesn't refresh the token if outside the refresh timeframe", () => {
```

src/usecases/refreshToken.test.js
```diff
@@ -0,0 +1,52 @@
...
+    };
+    const {
+      refreshedToken,
+      refreshedEncodedToken,
+      isTokenRefreshed,
+    } = refreshToken(container)(authenticationToken);
+
+    expect(isTokenRefreshed).toEqual(false);
+    expect(refreshedToken).toBeUndefined;
+    expect(refreshedEncodedToken).toBeUndefined;
```
chubberlisk, line 26: With our other usecases we still return the key but `null`. What do you think about following suit here?

src/usecases/verifyAdminToken.test.js
```diff
@@ -10,6 +10,7 @@ describe("verifyAdminToken", () => {
   const res = {
     writeHead: jest.fn(() => res),
     end: jest.fn(),
+    setHeader: jest.fn(),
```
chubberlisk, line 13: What do you think about adding a test to check that `verifyAdminToken` calls `refreshToken`?

src/usecases/refreshToken.js
jivdhaliwal: Date is initialised with milliseconds since Epoch, but the expiry on the token is stored in seconds since Epoch, so we need to multiply by 1000 here. 

I'm going to either move the magic number to a constant to be a bit more descriptive, or move to moment with has an initialiser that uses seconds

chubberlisk: Ah I see. Yush, that would be good. Thanks!

src/usecases/refreshToken.js
jivdhaliwal: Nope 👍

src/usecases/refreshToken.js
jivdhaliwal: Replaced it with moment.unix()

PR #362, opened 2020-05-29
Only show visits for the current trust on trust overview metrics

mackella:
# What

Update retrieveWardVisitTotals to accept an optional trustId parameter and modify trust-admin to pass the trustId.

# Why

The function will now return data for all trusts if no trustId is passed or filter the visit data to include only that for the specified trust.

# Screenshots

# Notes

### DIFF
--- a/src/usecases/retrieveWardVisitTotals.js
+++ b/src/usecases/retrieveWardVisitTotals.js
@@ -17,7 +17,8 @@ const getVisitsByWard = async (db, trustId) => {
     wards.hospital_id as hospital_id,
     wards.trust_id as trust_id
     FROM ward_visit_totals AS totals JOIN wards ON wards.id = totals.ward_id AND ($1 IS NULL OR wards.trust_id = $1)
-    GROUP BY wards.trust_id, wards.hospital_id, wards.name`,
+    GROUP BY wards.trust_id, wards.hospital_id, wards.name
+    ORDER BY wards.name`,
     [trustId]
   );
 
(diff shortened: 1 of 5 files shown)


src/usecases/retrieveWardVisitTotals.test.js
```diff
@@ -37,8 +65,51 @@ describe("retrieveWardVisitTotals", () => {
...
+      ]);
+
+      const container = {
+        getDb: async () => ({
+          any: anySpy,
+          trustId: 1,
+        }),
+      };
+
+      let res = await retrieveWardVisitTotals(container)();
```
chubberlisk, line 101: I think we need to pass in the `trustId` here / should explicitly do it in the test.

```suggestion
      const trustId = 1;
      let res = await retrieveWardVisitTotals(container)(trustId);
```

chubberlisk:
Picking up this PR and will add in a contract test.

src/usecases/retrieveWardVisitTotals.js
```diff
@@ -1,35 +1,41 @@
...
+        name
+      FROM
+        hospitals
+      WHERE
+        id = wards.hospital_id
+    ) as hospital_name,
+    wards.hospital_id as hospital_id,
+    wards.trust_id as trust_id
+    FROM ward_visit_totals AS totals JOIN wards ON wards.id = totals.ward_id AND ($1 IS NULL OR wards.trust_id = $1)
+    GROUP BY wards.trust_id, wards.hospital_id, wards.name
```
jivdhaliwal, line 20: Should we be grouping by `wards.name` here or `wards.id` ?  
Although likely rare, could we end up grouping two different wards with the same name (since it's not a unique field) ?

chubberlisk: Nice spot! 🕵️‍♂️  Will update.

PR #364, opened 2020-06-01
Move the Add a ward button on the view hospital page

jivdhaliwal:
# What
Moves the Add a ward button to the bottom of the wards list

# Why
Second iteration of changes to the view hospital page

# Screenshots
### Before
localhost_3000_trust-admin_hospitals_1(iPad) (1)
### After
localhost_3000_trust-admin_hospitals_1(iPad)

# Notes

### DIFF
--- a/src/components/Select/index.test.js
+++ b/src/components/Select/index.test.js
@@ -3,7 +3,7 @@ import { render, screen } from "@testing-library/react";
 import Select from "./index";
 
 describe("Select", () => {
-  it("renders the default option when no defaultValue is set", () => {
+  it("renders the prompt as the default option when no defaultValue is set", () => {
     render(
       <Select
         prompt="Choose an option"
@@ -17,7 +17,7 @@ describe("Select", () => {
     expect(screen.getByText("Choose an option")).toBeVisible();
   });
 
-  it("renders the default option when defaultValue is null", () => {
+  it("renders the prompt as the default option when defaultValue is null", () => {
     render(
       <Select
         defaultValue={null}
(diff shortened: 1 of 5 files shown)


src/components/Select/index.test.js
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import Select from "./index";
+
+describe("Select", () => {
+  it("renders the default option when no defaultValue is set", () => {
```
chubberlisk, line 6: This was a bit confusing for me with the use of "default option", I think what you're saying is:

```suggestion
  it("renders the prompt as the default option when no defaultValue is set", () => {
```

src/components/Select/index.test.js
```diff
@@ -0,0 +1,49 @@
...
+          { id: 1, name: "First Option" },
+          { id: 2, name: "Second Option" },
+        ]}
+      />
+    );
+
+    expect(screen.getByText("Choose an option")).toBeVisible();
+  });
+
+  it("renders the default option when defaultValue is null", () => {
```
chubberlisk, line 20: Same as above maybe?

src/components/Select/index.test.js
```diff
@@ -0,0 +1,49 @@
...
+          { id: 1, name: "First Option" },
+          { id: 2, name: "Second Option" },
+        ]}
+      />
+    );
+
+    expect(screen.getByText("Choose an option")).toBeVisible();
+  });
+
+  it("renders the the correct option when the defaultValue is set", () => {
```
chubberlisk, line 35: What is the "correct option"? Maybe something along the lines of:

```suggestion
  it("renders the default option as the option with the matching id as the defaultValue", () => {
```

PR #365, opened 2020-06-01
Add hospitals page for a trust admin

chubberlisk:
# What

Add a separate page for hospitals for a trust admin.

# Why

So we can have a single place where a trust admin can manage hospitals.

# Screenshots

image

# Notes

Next step is to show the booked visits in the table.

### DIFF
--- a/pages/trust-admin/hospitals/index.js
+++ b/pages/trust-admin/hospitals/index.js
@@ -1,0 +1,62 @@
+import React from "react";
+import Error from "next/error";
+import Layout from "../../../src/components/Layout";
+import propsWithContainer from "../../../src/middleware/propsWithContainer";
+import verifyTrustAdminToken from "../../../src/usecases/verifyTrustAdminToken";
+import HospitalsTable from "../../../src/components/HospitalsTable";
+import { GridRow, GridColumn } from "../../../src/components/Grid";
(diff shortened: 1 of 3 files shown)


pages/trust-admin/hospitals/index.js
```diff
@@ -0,0 +1,62 @@
...
+import verifyTrustAdminToken from "../../../src/usecases/verifyTrustAdminToken";
+import HospitalsTable from "../../../src/components/HospitalsTable";
+import { GridRow, GridColumn } from "../../../src/components/Grid";
+import Heading from "../../../src/components/Heading";
+import ActionLink from "../../../src/components/ActionLink";
+import Text from "../../../src/components/Text";
+
+const TrustAdmin = ({ hospitals, hospitalError, trust, trustError }) => {
+  if (hospitalError || trustError) {
+    return <Error />;
```
jivdhaliwal, line 14: If any errors occur, I believe this will hide them because we're not passing the error to next.js and then on to Sentry. `<Error err={hospitalError || trustError} />` (or passing a single error object as props) will do the trick (See _error.js)

chubberlisk: Updated as suggested. Thanks!

PR #370, opened 2020-06-02
Add booked visit totals to the view hospital page

jivdhaliwal:
# What
Adds the total booked visits to the view hospital page

# Why
Helps breakdown the usage across hospitals

# Screenshots
localhost_3000_trust-admin_hospitals_2(iPad)

# Notes

### DIFF
--- a/pageTests/trust-admin/hospitals/[id].test.js
+++ b/pageTests/trust-admin/hospitals/[id].test.js
@@ -49,6 +49,7 @@ describe("trust-admin/hospitals/[id]", () => {
       const container = {
         getRetrieveWardsByHospitalId: () => wardsSpy,
         getRetrieveHospitalById: () => hospitalSpy,
+        getRetrieveHospitalVisitTotals: () => jest.fn().mockReturnValue({}),
         getTokenProvider: () => tokenProvider,
       };
 
--- a/pageTests/trust-admin/hospitals/[id].test.js
+++ b/pageTests/trust-admin/hospitals/[id].test.js
@@ -14,8 +14,10 @@ describe("trust-admin/hospitals/[id]", () => {
   };
   let res;
 
+  const trustId = 1;
+
   const tokenProvider = {
-    validate: jest.fn(() => ({ type: "trustAdmin", trustId: 1 })),
+    validate: jest.fn(() => ({ type: "trustAdmin", trustId: trustId })),
   };
 
   beforeEach(() => {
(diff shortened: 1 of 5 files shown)


src/usecases/retrieveHospitalVisitTotals.js
```diff
@@ -0,0 +1,22 @@
...
+      JOIN wards ON wards.id = totals.ward_id
+      JOIN hospitals ON wards.hospital_id = hospitals.id
+    WHERE hospitals.trust_id = $1
+    GROUP BY hospitals.id`,
+    [trustId]
+  );
+  let response = {};
+
+  queryResult.map(({ hospital_id, total_visits }) => {
+    response[hospital_id] = total_visits;
```
chubberlisk, line 18: In `src/usecases/retrieveWardVisitTotals.js`, we parse the number to as an int. Could we do the same here to be consistent?

pageTests/trust-admin/hospitals/[id].test.js
```diff
@@ -49,6 +49,7 @@ describe("trust-admin/hospitals/[id]", () => {
       const container = {
         getRetrieveWardsByHospitalId: () => wardsSpy,
         getRetrieveHospitalById: () => hospitalSpy,
+        getRetrieveHospitalVisitTotals: () => jest.fn().mockReturnValue({}),
```
Comment (chubberlisk) [inline pageTests/trust-admin/hospitals/[id].test.js:52]: Could we add an expectation somewhere to check we're calling this usecase with the trust ID and that it gets set as props?

src/usecases/retrieveHospitalVisitTotals.contractTest.js
```diff
@@ -0,0 +1,83 @@
+import AppContainer from "../containers/AppContainer";
+
+describe("retrieveHospitalVisitTotals contract tests", () => {
+  const container = AppContainer.getInstance();
+
+  it("returns the total number of visits for each hospital", async () => {
+    const { trustId } = await container.getCreateTrust(container)({
```
chubberlisk, line 7: Do we need the `container` to be passed into `getCreateTrust`?

pageTests/trust-admin/hospitals/[id].test.js
Comment (jivdhaliwal) [reply pageTests/trust-admin/hospitals/[id].test.js:52]: Ye good point 👍

src/usecases/retrieveHospitalVisitTotals.contractTest.js
jivdhaliwal: good catch thanks

PR #371, opened 2020-06-02
Fix adding a ward showing an error when default hospital option

chubberlisk:
# What

Fix the add a ward form from showing the `Select a hospital` error when there's a default hospital option passed through.

# Why

This would be confusing for a user seeing an error to select a hospital when a hospital **_is_** selected.

# Screenshots

## Before

image

# Notes

Would like feedback on the test added, does it make sense?

### DIFF
--- a/pageTests/trust-admin/add-a-ward.test.js
+++ b/pageTests/trust-admin/add-a-ward.test.js
@@ -1,4 +1,7 @@
+import React from "react";
 import { getServerSideProps } from "../../pages/trust-admin/add-a-ward";
+import AddAWard from "../../pages/trust-admin/add-a-ward";
+import { render, screen, fireEvent } from "@testing-library/react";
 
 describe("/trust-admin/add-a-ward", () => {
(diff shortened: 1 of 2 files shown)


pageTests/trust-admin/add-a-ward.test.js
```diff
@@ -24,4 +27,22 @@ describe("/trust-admin/add-a-ward", () => {
...
+
+  describe("AddAWard", () => {
+    it("does not throw a hospital error when default select hospital option", () => {
+      render(
+        <AddAWard
+          hospitals={[
+            { id: "1", name: "Adora Hospital" },
+            { id: "2", name: "Catra Hospital" },
+          ]}
+          hospitalId={1}
```
jivdhaliwal, line 39: Does this need to be `defaultHospitalId` ?

jivdhaliwal: Oh never mind this is the AddAWard page, not the AddWardForm

PR #376, opened 2020-06-03
Add number of booked visits on trust admin hospitals page

chubberlisk:
# What

Add the number of booked visits for each hospital in the "Hospitals" page for a trust admin.

# Why

So they can compare usage between each of their hospitals.

# Screenshots

| Before | After |
|--------|-------|
|image|image|

# Notes

Next step is to show the number of wards in this table.

### DIFF
--- a/pages/trust-admin/hospitals/index.js
+++ b/pages/trust-admin/hospitals/index.js
@@ -47,10 +47,20 @@ export const getServerSideProps = propsWithContainer(
     const trustResponse = await container.getRetrieveTrustById()(
       authenticationToken.trustId
     );
+    const hospitalVisitTotalsResponse = await container.getRetrieveHospitalVisitTotals()(
+      authenticationToken.trustId
+    );
+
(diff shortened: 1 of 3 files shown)


pages/trust-admin/hospitals/index.js
```diff
@@ -9,11 +9,22 @@ import Heading from "../../../src/components/Heading";
...
+  hospitalError,
+  trust,
+  trustError,
+  hospitalVisitTotals,
+}) => {
   if (hospitalError || trustError) {
     return <Error err={hospitalError || trustError} />;
   }
 
+  const hospitalsWithVisitTotals = hospitals.map((hospital) => {
```
jivdhaliwal, line 23: What do you think about moving this to `getServerSideProps` so that we can test that `hospital.bookedVisits` is set correctly ?

chubberlisk: YES!

PR #378, opened 2020-06-03
Add number of wards for each hospital on trust admin hospitals page

chubberlisk:
# What

Show the number of wards for each hospital on the trust admin's hospitals page.

# Why

So that a trust admin can place the number of booked visits into perspective when comparing hospitals. 

# Screenshots

| Before | After |
|--------|-------|
|image|image|

# Notes

Not too sure about how I've added `withWards`. I wanted to avoid having to update wherever `retrieveHospitalsByTrustId` usecase is called.


src/usecases/retrieveHospitalsByTrustId.js
```diff
@@ -1,4 +1,7 @@
-const retrieveHospitalsByTrustId = ({ getDb }) => async (trustId) => {
+const retrieveHospitalsByTrustId = ({ getDb }) => async (
+  trustId,
+  options = {}
```
jivdhaliwal, line 3: Could we set the default withWards value here? That way it's easy to see what options this usecase takes without digging into the rest of the code

chubberlisk: Ah yush!

PR #388, opened 2020-06-04
Add Cypress test for booking a virtual visit

chubberlisk:
# What

Add a Cypress test for booking a virtual visit.

# Why

To increase our test coverage around a core user need of the service.

# Screenshots

N/A

# Notes

Would like feedback on:

- The way this test was written - wrapping functions Cypress calls in `Given`, `When`, `Then` functions.
- Placing the test file in a `wards` directory and thereby expecting other ward related tests to be in the directory.
- Whether the right amount of things has been tested. E.g. should the test click any of the links at the end of booking a visit page? Is testing the change links necessary? Is there anything missing?
- I added the `SHOW_NAVIGATION_BAR` through CircleCI interface, is that the right way to go? I couldn't get it to work using the `circle.yml` config file.


cypress/integration/wards/bookingAVisit.js
```diff
@@ -0,0 +1,97 @@
+describe("As a ward staff, I want to schedule a virtual visit so that patients can speak with their loved ones.", () => {
+  before(() => {
+    // reset and seed the database
+    cy.exec(
+      "npm run dbmigratetest reset && npm run dbmigratetest up && npm run db:seed"
+    );
+  });
+
+  it("allows a ward staff to book a virtual visit", () => {
```
jivdhaliwal, line 9: What do we think about splitting this test up into the form validations and the happy path (booking a visit without issues)? That way we can see if there are issues with just the form validation or vice versa without the other expectation getting in the way

cypress/integration/wards/bookingAVisit.js
```diff
@@ -0,0 +1,97 @@
...
+    WhenIFillOutForm();
+    AndISubmitTheForm();
+    ThenISeeTheCheckYourAnswersPage();
+
+    WhenIClickChange();
+    ThenISeeTheBookAVirtualVisitForm();
+
+    WhenISubmitTheForm();
+    AndIClickBookAVirtualVisit();
+    ThenISeeTheVirtualVisitIsBooked();
```
jivdhaliwal, line 29: Should we also check that the visit appears in the visits list ?

PR #395, opened 2020-06-08
Update logging in Cypress test to given, when, then format

chubberlisk:
# What

Update the logging in Cypress test to Given, When, Then format.

# Why

For readability and to be consistent with https://github.com/madetech/nhs-virtual-visit/blob/master/cypress/integration/wards/bookingAVisit.js.

# Screenshots

N/A

# Notes

N/A

### DIFF
--- a/cypress/integration/loggingIn.js
+++ b/cypress/integration/loggingIn.js
@@ -1,48 +1,100 @@
-describe("Logging in", () => {
-  describe("Given a valid ward code", () => {
-    before(() => {
-      // reset and seed the database
-      cy.exec(
-        "npm run dbmigratetest reset && npm run dbmigratetest up && npm run db:seed"
-      );
-    });
-    it("when a user enters this information, the user should be able to sign in", () => {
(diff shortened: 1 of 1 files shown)


cypress/integration/loggingIn.js
```diff
@@ -1,48 +1,100 @@
...
+  before(() => {
+    // reset and seed the database
+    cy.exec(
+      "npm run dbmigratetest reset && npm run dbmigratetest up && npm run db:seed"
+    );
   });
 
-  it("Given an invalid code, when a user enters this informtion, the user should see an error", () => {
+  it("allows a ward staff to log in and displays their home page", () => {
+    GivenIAmAWardStaff();
```
jivdhaliwal, line 10: Is this line required (given the function is empty) ?

chubberlisk: No, it's not required. It's there for readability. If preferred, I can remove?

jivdhaliwal: It keeps things consistent so I'm ok with leaving it in

PR #401, opened 2020-06-09
Add docker compose file for postgres, with setup files, and put instructions in dev readme

Robware:
# What
Adds scripts and instructions for setting up a local dev DB using docker

# Why
It removes the need for installing postgres locally

# Screenshots

# Notes

### DIFF
--- a/bin/setup_dev_db_docker.sh
+++ b/bin/setup_dev_db_docker.sh
@@ -3,4 +3,4 @@
 sudo docker-compose exec postgres dropdb nhs-virtual-visit-dev -U postgres
 sudo docker-compose exec postgres createdb nhs-virtual-visit-dev -U postgres
 npm run dbmigrate up
-sudo docker-compose exec postgres psql postgresql://postgres:postgres@localhost/nhs-virtual-visit-dev -f /db/seeds.sql
+sudo docker-compose exec postgres psql nhs-virtual-visit-dev -U postgres -f /db/seeds.sql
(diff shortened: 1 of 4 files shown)


docs/development/README.md
```diff
@@ -46,9 +46,25 @@ ENABLE_SENTRY=
 
 ### PostgreSQL 12
 
-If you are installing a local instance of PostgreSQL on Ubuntu, see the steps in the separate [guide.](./UBUNTU2004-PG12.md)
+#### Setup the database with docker
+
+Run `sudo docker-compose up --no-start` to create the container.
```
jivdhaliwal, line 51: Do we need to run this all as root ? (On OSX at least it ran ok without)

Robware: Currently, yes. Rootless mode is available but is experimental and limited.

bin/setup_dev_db_docker.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+sudo docker-compose exec postgres dropdb nhs-virtual-visit-dev -U postgres
+sudo docker-compose exec postgres createdb nhs-virtual-visit-dev -U postgres
+npm run dbmigrate up
+sudo docker-compose exec postgres psql postgresql://postgres:postgres@localhost/nhs-virtual-visit-dev -f /db/seeds.sql
```
jivdhaliwal, line 6: Is there anyway way we could avoid duplicating the DATABASE_URL used here and in .env ?

What if we shortened the line to only define the database name and the user ?
(e.g. `sudo docker-compose exec postgres psql -U postgres nhs-virtual-visit-dev` )

Or do we think it's fine to keep it as is since people won't be likely to use a DATABASE_URL different to the one in the readme when using docker ?

PR #402, opened 2020-06-09
Add Cypress test for starting a virtual visit

chubberlisk:
# What

Add Cypress test for starting a virtual visit.

# Why

To increase our test coverage around a core user need of the service.

# Screenshots

N/A

# Notes

Would like feedback/help with testing with the Whereby. It's currently kept to a minimal and just asserts for the `src`.


cypress/integration/wards/startingAVisit.js
```diff
@@ -0,0 +1,77 @@
...
+
+  function WhenIClickOnAttendVisit() {
+    cy.get("button").contains("Attend visit").click();
+  }
+
+  function ThenISeeAVideoFrame() {
+    cy.get("iframe").should(
+      "have.attr",
+      "src",
+      "https://nhs-virtual-visit.whereby.com/123?embed&iframeSource=nhs-virtual-visit&background=off&displayName=Ward&screenshare=off&chat=off"
```
jivdhaliwal, line 58: What do we think about adding a new, "test" provider, to the visits page and test the presence of that?

We have page tests that check that the Jitsi and Whereby frames are setup when the right provider is passed in, and since the whereby call will not actually connect in this test, it raises the question of what we are actually testing here. If it's just that there is a page with an iframe and an end call button then would it help to add a test provider option that does nothing except show an empty iframe?

It also means that we won't need to have a `WHEREBY_SUBDOMAIN` env set with the correct value for this test to run

chubberlisk: Having a look `pageTests/visits/[id].test.js` and as far as I can see it just tests Jitsi and not Whereby?

I think I feel a bit uneasy about adding code to the page just for tests. I believe we'd have to add another branch for the if statement and then replicate the button to do this. I'd prefer to work with what we have as prod code, so I'd propose removing checking the `src` as that's a bit spooky and not very user facing and just checking an iFrame and end call button exists (sort of done as we click on the button but better to be explicit maybe). 

Or maybe we just need to update the page test with Whereby coverage and remove the iFrame `src` expectation?

What do you think?

jivdhaliwal: >  removing checking the src as that's a bit spooky and not very user facing and just checking an iFrame and end call button exists

This sounds good to me 👍 If we want to check the src url is correct we could extract the whereby component out of the page to make it easier to test

PR #403, opened 2020-06-10
Trust admin usage stats

mackella:
# What

Added lists for hospitals with the most and least visits to the trust admin page.

# Why

To improve the metrics available to admin users.

# Screenshots

image

# Notes

retrieveHospitalVisitTotals has been modified so that it returns a list of hospitals ordered by visit count and also now includes the hospital names.

trustAdmin tests have been refactored to reduce duplication of setup code.


pages/trust-admin.js
```diff
@@ -86,6 +133,8 @@ export const getServerSideProps = propsWithContainer(
         wardError: wardsResponse.error,
         trustError: trustResponse.error,
         visitsScheduled: retrieveWardVisitTotals.total,
+        leastUsage: retrieveHospitalVisitTotals.slice(0, 3),
```
jivdhaliwal: What do we think about moving this logic out of the page and into the usecase (similar to https://github.com/madetech/nhs-virtual-visit/blob/fd9bd2c358630079f265669a2a2370814321b8f1/src/usecases/retrieveHospitalWardVisitTotals.js#L44
That way the page is not making assumptions about the logic in retrieveHospitalVisitTotals (i.e. the totals are returned in order of total_visits DESC)

mackella: Yes, good point, I think that would be better. I hadn't spotted the ward implementation!

mackella: Fixed now and hyperlinks added.

PR #419, opened 2020-06-12
 Add edit journey (missing success page)

Robware:
# What
UI to edit a hospital
# Why
The user needs to edit a hospital

# Screenshots
Screen Shot 2020-06-12 at 10 17 46

Screen Shot 2020-06-12 at 10 17 51

# Notes
Success page still pending

### DIFF
--- a/cypress/integration/trust-admin/editingAHospital.js
+++ b/cypress/integration/trust-admin/editingAHospital.js
@@ -1,0 +1,61 @@
+describe("As an admin, I want to edit a hospital so that I can keep hospital changes up to date.", () => {
+  before(() => {
+    // reset and seed the database
+    cy.exec(
+      "npm run dbmigratetest reset && npm run dbmigratetest up && npm run db:seed"
+    );
+  });
+
+  function GivenIAmLoggedInAsAnAdmin() {
+    cy.visit(Cypress.env("baseUrl") + "/trust-admin/login");
+    cy.get("input[name=code]").type(Cypress.env("validTrustAdminCode"));
+    cy.get("input[name=password]").type(Cypress.env("validTrustAdminPassword"));
+    cy.get("button").contains("Log in").click();
+  }
+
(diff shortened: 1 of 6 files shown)


src/components/HospitalsTable/index.js
```diff
@@ -28,6 +28,16 @@ const HospitalsTable = ({ hospitals }) => (
...
+              <AnchorLink
+                href={`/trust-admin/hospitals/${hospital.id}/edit`}
+                className="nhsuk-u-margin-right-4"
+              >
+                Edit
+                <span className="nhsuk-u-visually-hidden">
+                  {" "}
+                  {hospital.name}
+                </span>
+              </AnchorLink>
```
chubberlisk, line 40: Could we make it so the `View` link appears first? So we're consistent with the wards table so the actions links on table go view, edit and delete.

Robware: Good spot, done.

cypress/integration/trust-admin/editingAHospital.js
```diff
@@ -0,0 +1,82 @@
+describe("As an admin, I want to edit a hospital so that I can keep hospital changes up to date.", () => {
+  beforeEach(() => {
```
jivdhaliwal, line 2: Do we need to run this setup before each test in this file? Ideally we want to run this as little as possible to reduce the runtime of the e2e tests

Robware: If I don't then `ThenIShouldBeOnTheEditHospitalPage` doesn't work the second time round. I can update the tests to work around this, if you'd prefer?

jivdhaliwal: Ah in that case what if we re-ordered the specs so that we didn't have to run the setup twice ?

Robware: Good idea, wish I'd thought of that.

PR #420, opened 2020-06-12
Event capture workflow

mackella:
# What

Added an event logging system, which will initially be used to capture visit participation events, e.g. joining and leaving a video call.

# Why

To support additional visit metrics and to provided a framework for capturing additional metrics in future.

# Screenshots

N/A

# Notes

* capture-event API and use case created
* events table added
* status added to scheduled_calls_table (scheduled|cancelled|archived|complete)
* clean up script modified to remove personal data instead of deleting records
* affected db queries modified to take status into account

### DIFF
--- a/db/migrations/sqls/20200612085103-add-visit-status-up.sql
+++ b/db/migrations/sqls/20200612085103-add-visit-status-up.sql
@@ -1,1 +1,3 @@
-ALTER TABLE scheduled_calls_table ADD status varchar(20);
+ALTER TABLE scheduled_calls_table ADD status varchar(20) NOT NULL DEFAULT 'scheduled';
+
+UPDATE scheduled_calls_table SET status = 'scheduled';
--- a/db/migrations/sqls/20200612085103-add-visit-status-up.sql
+++ b/db/migrations/sqls/20200612085103-add-visit-status-up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE scheduled_calls_table ADD status varchar(20);
--- a/db/scripts/cleanup_scheduled_calls.js
+++ b/db/scripts/cleanup_scheduled_calls.js
@@ -13,7 +13,9 @@ async function getDb() {
 async function cleanupScheduledCalls() {
   const db = await getDb();
   const scheduledCalls = await db.result(
-    "DELETE from scheduled_calls_table WHERE call_time < (now() - INTERVAL '1 DAY')"
+    `UPDATE scheduled_calls_table 
+     SET patient_name = null, recipient_number = null, recipient_name = null, call_password = null, recipient_email = null, status = 'complete'
+     WHERE call_time < (now() - INTERVAL '1 DAY')`
   );
   return scheduledCalls;
 }
--- a/db/scripts/cleanup_scheduled_calls.js
+++ b/db/scripts/cleanup_scheduled_calls.js
@@ -1,3 +1,5 @@
+import { COMPLETE } from "../../src/helpers/visitStatus";
+
 async function getDb() {
   const dotenv = require("dotenv");
   dotenv.config();
@@ -14,8 +16,9 @@ async function cleanupScheduledCalls() {
   const db = await getDb();
   const scheduledCalls = await db.result(
     `UPDATE scheduled_calls_table 
-     SET patient_name = null, recipient_number = null, recipient_name = null, call_password = null, recipient_email = null, status = 'complete'
-     WHERE call_time < (now() - INTERVAL '1 DAY')`
+     SET patient_name = null, recipient_number = null, recipient_name = null, call_password = null, recipient_email = null, status = $1
+     WHERE call_time < (now() - INTERVAL '1 DAY')`,
+    COMPLETE
   );
   return scheduledCalls;
 }
--- a/db/scripts/seed_database.js
+++ b/db/scripts/seed_database.js
@@ -24,8 +24,8 @@ async function seedDatabase() {
 
   await db.result(
     `INSERT INTO scheduled_calls_table
-    (patient_name, recipient_email, recipient_name, call_time, call_id, provider, ward_id, call_password)
-    VALUES ('Alice', 'bob@example.com', 'Bob', CURRENT_TIMESTAMP, '123', 'whereby', $1, 'password')`,
+    (patient_name, recipient_email, recipient_name, call_time, call_id, provider, ward_id, call_password, status)
+    VALUES ('Alice', 'bob@example.com', 'Bob', CURRENT_TIMESTAMP, '123', 'whereby', $1, 'password', 'scheduled')`,
     [wardId]
   );
   db.$pool.end();
--- a/db/scripts/seed_database.js
+++ b/db/scripts/seed_database.js
@@ -28,7 +28,7 @@ async function seedDatabase() {
   await db.result(
     `INSERT INTO scheduled_calls_table
     (patient_name, recipient_email, recipient_name, call_time, call_id, provider, ward_id, call_password, status)
-    VALUES ('Alice', 'bob@example.com', 'Bob', CURRENT_TIMESTAMP, '123', 'whereby', $1, 'password', 'scheduled')`,
+    VALUES ('Alice', 'bob@example.com', 'Bob', CURRENT_TIMESTAMP + interval '1 hour', '123', 'whereby', $1, 'password', 'scheduled')`,
     [wardId]
   );
   db.$pool.end();
--- a/src/helpers/visitStatus.js
+++ b/src/helpers/visitStatus.js
@@ -1,0 +1,6 @@
+// Defined the set of statuses that a visit can have (status column in scheduled_calls_table)
+
+export const SCHEDULED = "scheduled"; // Default - initial state
+export const ARCHIVED = "archived"; // The visit's parent ward has been archived
+export const CANCELLED = "cancelled"; // The visit has been explicitly cancelled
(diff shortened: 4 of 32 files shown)


src/usecases/retrieveVisits.js
```diff
@@ -1,7 +1,9 @@
 const retrieveVisits = ({ getDb }) => async ({ wardId }) => {
   const db = await getDb();
   try {
-    let query = `SELECT * FROM scheduled_calls_table WHERE ward_id = $1 ORDER BY call_time ASC`;
+    let query = `SELECT * FROM scheduled_calls_table 
+                 WHERE ward_id = $1 AND call_time >= NOW() AND status = 'scheduled'
```
jivdhaliwal, line 5: We filter the visits by callTime in the `AccordionVisits` component (using 3 filter helpers). As such we don't need to check for visits that have a call time in the future (`call_time >= NOW()`). This query will prevent us from showing visits from the last 12 hours (one of the filters on the Ward visits page)

jivdhaliwal: This change is also the reason the e2e tests are failing. A visit is scheduled with a call_time of now, and then the test runs a second after and filters out these 'past' visit.

db/scripts/cleanup_scheduled_calls.js
```diff
@@ -13,7 +13,9 @@ async function getDb() {
 async function cleanupScheduledCalls() {
```
jivdhaliwal, line 13: To avoid any confusion, could we delete the unused `db/cleanup_old_scheduled_calls.sql` ?

db/migrations/sqls/20200612085103-add-visit-status-up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE scheduled_calls_table ADD status varchar(20);
```
jivdhaliwal, line 1: What do we think about defaulting the status to 'scheduled' in the schema?

jivdhaliwal, line 1: How do we plan to deal with the existing visits in the scheduled_calls_table ? Should we update this migration to also update the status on any existing calls to be 'scheduled' (as all the completed visits will have been deleted)

src/usecases/archiveWard.test.js
```diff
@@ -35,7 +35,7 @@ describe("archiveWard", () => {
       error: null,
     });
     expect(resultSpy).toHaveBeenCalledWith(
-      expect.stringMatching(/delete from scheduled_calls_table/i),
+      expect.stringMatching(/UPDATE scheduled_calls_table/i),
```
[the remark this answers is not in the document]
mackella: Done!

src/usecases/archiveWard.js
```diff
@@ -16,10 +16,9 @@ const archiveWard = ({ getRetrieveWardById, getDb }) => async (
 
   try {
     await db.result(
-      `DELETE FROM scheduled_calls_table
-        WHERE
-            ward_id = $1
-    `,
+      `UPDATE scheduled_calls_table
+       SET status = 'archived'
```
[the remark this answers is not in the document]
mackella: Constants added, as per Slack discussion.

src/usecases/retrieveVisits.js
mackella: I've removed the time constraint from that query and reviewed all usage without finding any negative consequences.

db/scripts/cleanup_scheduled_calls.js
mackella: Done.

db/migrations/sqls/20200612085103-add-visit-status-up.sql
mackella: Backfill query added to migration.

mackella: Done.

db/scripts/seed_database.js
```diff
@@ -27,8 +27,8 @@ async function seedDatabase() {
 
   await db.result(
     `INSERT INTO scheduled_calls_table
-    (patient_name, recipient_email, recipient_name, call_time, call_id, provider, ward_id, call_password)
-    VALUES ('Alice', 'bob@example.com', 'Bob', CURRENT_TIMESTAMP, '123', 'whereby', $1, 'password')`,
+    (patient_name, recipient_email, recipient_name, call_time, call_id, provider, ward_id, call_password, status)
+    VALUES ('Alice', 'bob@example.com', 'Bob', CURRENT_TIMESTAMP + interval '1 hour', '123', 'whereby', $1, 'password', 'scheduled')`,
```
jivdhaliwal, line 31: I think we should keep the timestamp as the current time so that we have a visit that is starting 'now' in case the scheduled visit filtering code ends up affecting the visits tests again

mackella: Hmm.. not sure about that. My thinking is that the visit isn't starting 'now' from the PoV of the tests, but a few seconds/minutes ago. I think what we should really have is a set of visits defined to match each state so that we can test them all.
Definitely room for improvement though. Perhaps we could improve the seed data while doing one of the other metric tickets?

jivdhaliwal: A visit that started a few seconds ago represents a visit starting now in the real world (staff start visits well after they are meant to start which is why we keep visits around so long)

jivdhaliwal: I'm keen to unblock this PR though so happy to merge and fix the missing test coverage around retrieveVisits later

src/helpers/visitStatus.js
```diff
@@ -0,0 +1,6 @@
+// Defined the set of statuses that a visit can have (status column in scheduled_calls_table)
+
+export const SCHEDULED = "scheduled"; // Default - initial state
+export const ARCHIVED = "archived"; // The visit's parent ward has been archived
+export const CANCELLED = "cancelled"; // The visit has been explicitly cancelled
+export const ADMIN = "complete"; // The visit has occurred
```
chubberlisk, line 6:
```suggestion
export const COMPLETE = "complete"; // The visit has occurred
```

Looking at `db/scripts/cleanup_scheduled_calls.js`, it looks like we want this instead?🕵️‍♀️  Maybe it should be past tense like the others?

db/scripts/seed_database.js
chubberlisk, line 31: Is it worth using the `src/helpers/visitStatus.js` here?

src/usecases/createVisit.js
```diff
@@ -1,10 +1,12 @@
+import { SCHEDULED } from "../../src/helpers/visitStatus";
+
 const createVisit = ({ getDb }) => async (visit) => {
   const db = await getDb();
 
   return await db.one(
     `INSERT INTO scheduled_calls_table
-      (id, patient_name, recipient_email, recipient_number, recipient_name, call_time, call_id, provider, ward_id, call_password)
-      VALUES (default, $1, $2, $3, $4, $5, $6, $7, $8, $9)
+      (id, patient_name, recipient_email, recipient_number, recipient_name, call_time, call_id, provider, ward_id, call_password, status)
```
chubberlisk, line 8: Do we need this change anymore since we're defaulting the status value to `scheduled`?

db/scripts/cleanup_scheduled_calls.js
```diff
@@ -13,7 +15,10 @@ async function getDb() {
...
   const db = await getDb();
   const scheduledCalls = await db.result(
-    "DELETE from scheduled_calls_table WHERE call_time < (now() - INTERVAL '1 DAY')"
+    `UPDATE scheduled_calls_table 
+     SET patient_name = null, recipient_number = null, recipient_name = null, call_password = null, recipient_email = null, status = $1
+     WHERE call_time < (now() - INTERVAL '1 DAY')`,
+    COMPLETE
```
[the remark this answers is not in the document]
mackella: From observation of the rest of the codebase, it seems you can use either form if there's only a single value.
The docs at http://vitaly-t.github.io/pg-promise/Database.html#result seems to confirm this.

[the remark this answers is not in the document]
chubberlisk: Cooooool! Just double checking, wasn't sure myself.

db/scripts/seed_database.js
mackella: I tried to do that and it didn't like the import when running the npm db:seed task. Not sure why - perhaps ES6 isn't supported in that scenario. Any ideas?

src/helpers/visitStatus.js
mackella: That's a great spot and on reflection also explains why the db seeding was failing as I was referencing `COMPLETED` in that script :facepalm: 
I'll fix it...

db/scripts/seed_database.js
jivdhaliwal: correct this file isn't transpiled so you won't be able to pull in the constants

src/helpers/visitStatus.js
mackella: Actually, it doesn't explain the seeding issue as that was `status = SCHEDULED`...

src/usecases/createVisit.js
mackella: I guess not, although making it explicit is perhaps easier to follow from a maintenance perspective.

src/helpers/visitStatus.js
mackella: The name is fixed now...

PR #424, opened 2020-06-15
Add contract test for retrieveVisits usecase

jivdhaliwal:
# What
Adds a retrieveVisits tests that hits the test database

# Why
To ensure that the underlying query works as intended

# Screenshots

# Notes

### DIFF
--- a/src/usecases/retrieveVisits.contractTest.js
+++ b/src/usecases/retrieveVisits.contractTest.js
@@ -1,0 +1,76 @@
+import AppContainer from "../containers/AppContainer";
+import moment from "moment";
+import deleteVisitByCallId from "./deleteVisitByCallId";
+
+describe("retrieveVisits contract tests", () => {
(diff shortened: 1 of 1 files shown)


src/usecases/retrieveVisits.contractTest.js
```diff
@@ -0,0 +1,76 @@
+import AppContainer from "../containers/AppContainer";
+import moment from "moment";
+import deleteVisitByCallId from "./deleteVisitByCallId";
+
+describe("retrieveVisits contract tests", () => {
+  const container = AppContainer.getInstance();
+
+  it("retrieves all Trusts from the db", async () => {
```
mackella, line 8: Should that be 'visits' instead of 'trusts'?

jivdhaliwal: copy paste fail, nice catch

PR #427, opened 2020-06-16
Added contract test for cleanup script and modified db visit cleanup logic to better handle visit status

mackella:
# What

Modified db visit cleanup logic to better handle visit status
 - ensure all complete, archived and cancelled visits are redacted
 - only change visit status to complete when currently scheduled 

# Why

To ensure personal data is not retained in any visit record, regardless of status.

# Screenshots

N/A

# Notes

There is currently no easy way to retrieve a visit which is not in the `scheduled` status in tests. Should we implement this, either to support future functional requirements or purely for testing?

### DIFF
--- a/db/scripts/cleanup_scheduled_calls.js
+++ b/db/scripts/cleanup_scheduled_calls.js
@@ -5,15 +5,19 @@ async function getDb() {
   dotenv.config();
 
   const pgp = require("pg-promise")({});
-  const connectionURL =
-    process.env.NODE_ENV === "test" || process.env.APP_ENV === "test"
-      ? process.env.TEST_DATABASE_URL
-      : process.env.DATABASE_URL;
 
-  return pgp({
-    connectionString: connectionURL,
-    ssl: { rejectUnauthorized: process.env.NODE_ENV === "production" },
-  });
+  let options = {
+    connectionString:
+      process.env.NODE_ENV === "test" || process.env.APP_ENV === "test"
+        ? process.env.TEST_DATABASE_URL
+        : process.env.DATABASE_URL,
+  };
+
+  if (process.env.NODE_ENV === "production") {
+    options.ssl = { rejectUnauthorized: false };
+  }
+
+  return pgp(options);
 }
 
 async function cleanupScheduledCalls() {
--- a/db/scripts/cleanup_scheduled_calls.js
+++ b/db/scripts/cleanup_scheduled_calls.js
@@ -6,12 +6,13 @@ async function getDb() {
 
   const pgp = require("pg-promise")({});
   const connectionURL =
-    process.env.NODE_ENV === "test"
+    process.env.NODE_ENV === "test" || process.env.APP_ENV === "test"
       ? process.env.TEST_DATABASE_URL
       : process.env.DATABASE_URL;
+
   return pgp({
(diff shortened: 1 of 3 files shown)


db/scripts/cleanup_scheduled_calls.js
```diff
@@ -14,19 +17,44 @@ async function getDb() {
 
 async function cleanupScheduledCalls() {
   const db = await getDb();
```
jivdhaliwal, line 19: Now that this file is running as part of the test suite, we should close the connection once we're done so that there are no loose connections hanging around (like this:  https://github.com/madetech/nhs-virtual-visit/blob/df12a1ff523480b840aca16f282ccfba7143d2b4/testTeardown.js#L14-L17

We may need to pass the db in the result so that we can close it in a ["finally"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally) call at the end

db/scripts/cleanup_scheduled_calls.js
```diff
@@ -5,7 +5,10 @@ async function getDb() {
...
 
   const pgp = require("pg-promise")({});
-  const connectionURL = process.env.DATABASE_URL;
+  const connectionURL =
+    process.env.NODE_ENV === "test"
+      ? process.env.TEST_DATABASE_URL
+      : process.env.DATABASE_URL;
   return pgp({
     connectionString: connectionURL,
     ssl: { rejectUnauthorized: false },
```
jivdhaliwal, line 14: The test database on circleCI (and local dev) does not use an SSL connection, so we will need to update this to only set in production (like https://github.com/madetech/nhs-virtual-visit/blob/master/src/gateways/Database/index.js#L17-L19

db/scripts/cleanup_scheduled_calls.js
```diff
@@ -5,28 +5,59 @@ async function getDb() {
...
-  const connectionURL = process.env.DATABASE_URL;
+  const connectionURL =
+    process.env.NODE_ENV === "test" || process.env.APP_ENV === "test"
+      ? process.env.TEST_DATABASE_URL
+      : process.env.DATABASE_URL;
+
   return pgp({
     connectionString: connectionURL,
-    ssl: { rejectUnauthorized: false },
+    ssl: { rejectUnauthorized: process.env.NODE_ENV === "production" },
```
jivdhaliwal, line 15: The presence of the ssl key enables / disables SSL in the connection. You will need to remove the key entirely if you want to disable SSL (not just the rejectUnauthorized flag)

db/scripts/cleanup_scheduled_calls.js
mackella: I'd already done this, but in the script itself rather than the test teardown. I think it should be there because if the script is run manually it waits until it times out rather than completing immediately otherwise.

db/scripts/cleanup_scheduled_calls.js
mackella: Done. Seems to work now :slightly_smiling_face:

db/scripts/cleanup_scheduled_calls.js
mackella: Done

PR #428, opened 2020-06-16
Server side password hashing

jivdhaliwal:
# What
Hashes passwords in the application

# Why
Ensures passwords are not logged in any way (e.g. postgres query logs)

# Screenshots

# Notes


jivdhaliwal:
> These changes look fine!
> I have a question on the description/intent though. Surely this is still server-side hashing? I get that the db will no longer see the plaintext passwords, but they'll still be in server logs, etc. If we want client-based hashing we'd need to implement it wholly within the page.

ye good point, we actually meant to name this server side or app side hashing as that is actually what is going on 👍

PR #432, opened 2020-06-17
Update GovNotify gateway to return fake NotifyClient when test env

chubberlisk:
# What

Update `GovNotify` gateway to return the fake `NotifyClient` when test env and no longer skip the booking, starting and attending a visit Cypress tests. 

# Why

We added Cypress tests to cover booking, starting and attending a visit however, they were causing tests to be flakey because they call Notify and sometimes it would take longer than Cypress was willing to wait. As a result, they were skipped (https://github.com/madetech/nhs-virtual-visit/pull/421) to avoid CI from failing.

This PR means that when Cypress tests are run, we won't be calling Notify anymore and means that tests are more reliable.

# Screenshots

N/A

# Notes

- Best reviewed commit by commit.
- Doing `notifyClient.sendEmail = jest.fn(notifyClient.sendEmail);` feels a bit weird but feels like a fair compromise to allow us to use the fake Notify in the app. What do you think?

### DIFF
--- a/src/gateways/GovNotify/index.test.js
+++ b/src/gateways/GovNotify/index.test.js
@@ -1,4 +1,9 @@
+import { NotifyClient } from "notifications-node-client";
 import GovNotify from "./";
+
+const {
+  NotifyClient: FakeNotifyClient,
+} = require("../../../__mocks__/notifications-node-client");
 
 describe("GovNotify", () => {
   it("Produces a singleton", () => {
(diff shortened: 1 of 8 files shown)


src/gateways/GovNotify/index.test.js
```diff
@@ -10,4 +15,36 @@ describe("GovNotify", () => {
 
     expect(instanceOne).toEqual(instanceTwo);
   });
+
+  describe("when using production environment", () => {
+    beforeEach(() => {
+      process.env.NODE_ENV = "production";
```
jivdhaliwal, line 21: Since the code only looks at the APP_ENV, do we need to be changing the NODE_ENV here as it has no affect on the fake client logic? In production the APP_ENV is simply not defined

chubberlisk: Yeah, I went back and forth with this. Will remove!

src/gateways/GovNotify/index.test.js
```diff
@@ -10,4 +15,28 @@ describe("GovNotify", () => {
 
     expect(instanceOne).toEqual(instanceTwo);
   });
+
+  describe("when using production environment", () => {
```
jivdhaliwal, line 19:
```suggestion
  describe("when we are not running a test server", () => {
```
What do we think about changing the description to more closely match the behaviour ?

PR #440, opened 2020-06-18
Trigger visit events

mackella:
# What

Added event triggers for join-visit and leave-visit.

# Why

To provide source data for deriving usage metrics.

# Screenshots

N/A

# Notes

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Please take a look at our separate [Contributing Guide](./CONTRIBUTING.md).
 - **Joshua-Luke Bevan** - Software Engineer at [Made Tech](https://www.madetech.com) (joshua.bevan@madetech.com)
 - **Stephen Thomson** - Senior Software Engineer at [Made Tech](https://www.madetech.com) (stephen.thomson@madetech.com)
 - **Neil Kidd** - Lead Software Engineer at [Made Tech](https://www.madetech.com) (neil.kidd@madetech.com)
+- **Stu Mackellar** - Lead Software Engineer at [Made Tech](https://www.madetech.com) (stu.mackellar@madetech.com)
 
 ## License
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1483,6 +1483,12 @@
             "psl": "^1.1.28",
             "punycode": "^2.1.1"
           }
+        },
+        "uuid": {
+          "version": "3.4.0",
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -73,6 +73,7 @@ Please take a look at our separate [Contributing Guide](./CONTRIBUTING.md).
 - **Joshua-Luke Bevan** - Software Engineer at [Made Tech](https://www.madetech.com) (joshua.bevan@madetech.com)
 - **Stephen Thomson** - Senior Software Engineer at [Made Tech](https://www.madetech.com) (stephen.thomson@madetech.com)
 - **Neil Kidd** - Lead Software Engineer at [Made Tech](https://www.madetech.com) (neil.kidd@madetech.com)
+- **Stu Mackellar** - Lead Software Engineer at [Made Tech](https://www.madetech.com) (stu.mackellar@madettech.com)
```
chubberlisk, line 76: 🎉 

```suggestion
- **Stu Mackellar** - Lead Software Engineer at [Made Tech](https://www.madetech.com) (stu.mackellar@madetech.com)
```

pages/api/capture-event.js
```diff
@@ -28,7 +28,7 @@ export default withContainer(
       return;
     }
 
-    if (!body.visitId || isNaN(body.visitId)) {
+    if (body.visitId === null || body.visitId === "" || isNaN(body.visitId)) {
```
jivdhaliwal, line 31: Would this cover visitId being undefined?

jivdhaliwal, line 31: Could you clarify why we would want to accept 0 as a visitId ? (based on the message in a8ce8e9d7439c).

If we want to explicitly allow 0, have we considered including that in the statement?

(e.g. `if ((!body.visitId && body.visitId !== 0) || isNaN(body.visitId))`

This should also cover an undefined visitId

chubberlisk: I think at the time we reckoned 0 was a valid `visitId`. Now I'm wondering whether it is, looking at the DB?

jivdhaliwal: The default minimum value is 1 so we should never had visits with an ID of 0 (https://www.postgresql.org/docs/6.4/sql-createsequence.htm)

chubberlisk: Awesome, I'll drop the commit :)

chubberlisk: Yush, definitely now that we've dropped worrying about a 0 visitId.

PR #441, opened 2020-06-18
Add trust factory for testing

jivdhaliwal:
# What
This change introduces a new test factory, 'setupTrust', which allows us to quickly
setup a trust in a test without having to worry about making sure all the values
are set correctly

# Why
If we need a specific value for a test, one can be passed in to the factory to
override the defaults.

This allows us make changes to trust creation without having to change all the tests
(e.g. adding a new required field)

# Screenshots

# Notes


src/testUtils/fixtures/setupTrust.js
```diff
@@ -0,0 +1,14 @@
...
+export default async ({
+  name = null,
+  adminCode = null,
+  password = null,
+} = {}) => {
+  return await container.getCreateTrust()({
+    name: name || "Test Trust",
+    adminCode: adminCode || "TESTCODE",
+    password: password || "TESTPASSWORD",
+  });
```
antonyoneill, line 13: Nice. You can also apply a similar pattern here to prevent repeating keys
```javascript
export default sync (args = {}) => {
    return await container.getCreateTrust()({
        name: 'Test Trust',
        adminCode: 'TESTCODE',
        password: 'TESTPASSWORD',
        ...args
    });
};
```

Though not sure if you give up some intellisense hints doing that

PR #443, opened 2020-06-19
Add video provider to trusts

jivdhaliwal:
# What
Allows Admins to select a video provider when creating a new Trust. This change also updates all existing trusts to use the whereby provider as that is currently the default provider for all current trusts.

# Why
To give the Admins the choice of which video provider to use

# Screenshots

# Notes
This change focuses on storing the preference in the database. A later PR will hook the video provider preference up to the visit booking logic

### DIFF
--- a/cypress/integration/admin/addingATrust.js
+++ b/cypress/integration/admin/addingATrust.js
@@ -50,6 +50,7 @@ describe("As an admin, I want to add a trust so that a trust can use the virtual
     cy.get("input[name=trust-admin-code]").type("bowcode");
     cy.get("input[name=trust-password]").type("bowpassword");
     cy.get("input[name=trust-password-confirmation]").type("bowpassword");
+    cy.get("select").select("whereby");
   }
 
   function AndISubmitTheForm() {
--- a/cypress/integration/admin/addingATrust.js
+++ b/cypress/integration/admin/addingATrust.js
@@ -47,10 +47,10 @@ describe("As an admin, I want to add a trust so that a trust can use the virtual
 
   function WhenIFillOutTheForm() {
     cy.get("input[name=trust-name]").type("Bow Trust");
+    cy.get("select[name=video-provider]").select("whereby");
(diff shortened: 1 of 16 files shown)


pages/admin/add-a-trust.js
```diff
@@ -103,6 +117,10 @@ const AddATrust = () => {
       setPasswordConfirmationError(onSubmitErrors);
     }
 
+    if (videoProvider.length === 0) {
```
chubberlisk, line 120: Could we move this to after checking if there is an error in the name?

The reason being is the `Select a video provider` error message ends up in the bottom of the error summary. The error messages in the error summary should follow the order of the inputs on the page. It shucks that it's so tied to the presentation of it but not sure how to do it without a ton of refactoring. :/

image

cypress/integration/admin/addingATrust.js
```diff
@@ -50,6 +50,7 @@ describe("As an admin, I want to add a trust so that a trust can use the virtual
     cy.get("input[name=trust-admin-code]").type("bowcode");
     cy.get("input[name=trust-password]").type("bowpassword");
     cy.get("input[name=trust-password-confirmation]").type("bowpassword");
+    cy.get("select").select("whereby");
```
chubberlisk, line 53: Could we be more explicit here with the `cy.get`? E.g. `cy.get("select[name=video-provider]` so if a newcomer came along they'd know Whereby is a video provider and also if we added another dropdown we don't have to update this.

pages/api/create-trust.js
```diff
@@ -34,6 +34,12 @@ export default withContainer(
       return;
     }
 
+    if (!body.videoProvider) {
```
chubberlisk, line 37: I think we're missing a test for this. 🕵️‍♀️

cypress/integration/admin/addingATrust.js
jivdhaliwal: Good point. I was being a bit lazy with the select here, will update thanks

PR #445, opened 2020-06-22
Fix visits page not capturing events when a key contact

chubberlisk:
# What

Fix the visits page not capturing events when a key contact by updating the `api/capture-event` endpoint to allow authentication via the `callId` and `callPassword`.

# Why

Because we need events from a key contact's perspective for metrics.

# Screenshots

N/A

# Notes

N/a

### DIFF
--- a/pages/api/capture-event.js
+++ b/pages/api/capture-event.js
@@ -11,30 +11,18 @@ export default withContainer(
       return;
     }
 
-    if (headers?.cookie) {
-      const userIsAuthenticated = container.getUserIsAuthenticated();
-      const userIsAuthenticatedResponse = await userIsAuthenticated(
-        headers.cookie
-      );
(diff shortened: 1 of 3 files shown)


pages/api/capture-event.js
```diff
@@ -11,12 +11,30 @@ export default withContainer(
       return;
     }
 
-    const userIsAuthenticated = container.getUserIsAuthenticated();
-    const userIsAuthenticatedResponse = await userIsAuthenticated(
-      headers.cookie
-    );
+    if (headers?.cookie) {
```
jivdhaliwal, line 14: This makes the assumption that if they have a cookie set, then they should go through the normal authentication. If we were to store a cookie on key contact visitors, this may stop capturing their events in the future. 

What if we skip the `headers?.cookie` check and do both the user authentication and the call password verification, returning unauthorised if both return false?

PR #455, opened 2020-06-24
Add usecase to get the average number of participants in a visit

chubberlisk:
# What

Add usecase to get the average number of participants in a visit for a trust as well as small refactoring to use a helper for the event action names.

# Why

So we can display the metric on the trust admin dashboard.

# Screenshots

N/A

# Notes

Best reviewed commit by commit.

Would like feedback on:

- How do deal with the long contract test.
- Are there any edge cases I'm missing?

### DIFF
--- a/src/usecases/retrieveAverageParticipantsInVisit.js
+++ b/src/usecases/retrieveAverageParticipantsInVisit.js
@@ -1,0 +1,23 @@
+const retrieveAverageParticipantsInVisit = ({ getDb }) => async (trustId) => {
+  if (!trustId) return { error: "A trustId must be provided." };
+
+  const db = await getDb();
+  const [{ average_participants: averageParticipantsInVisit }] = await db.any(
+    `SELECT AVG(participants) as average_participants
+    FROM (SELECT COUNT ( DISTINCT session_id ) AS participants
+          FROM events
+          LEFT JOIN scheduled_calls_table ON events.visit_id = scheduled_calls_table.id
+          LEFT JOIN wards ON scheduled_calls_table.ward_id = wards.id
+          LEFT JOIN trusts ON wards.trust_id = trusts.id
+          WHERE trusts.id = $1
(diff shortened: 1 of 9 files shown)


src/usecases/retrieveAverageParticipantsInVisit.js
```diff
@@ -0,0 +1,54 @@
...
+
+    for (const [, sessionEvents] of Object.entries(eventsGroupedBySesssion)) {
+      const joinEvent = sessionEvents.find(
+        (event) => event.action === JOIN_VISIT
+      );
+      const leaveEvent = sessionEvents.find(
+        (event) => event.action === LEAVE_VISIT
+      );
+
+      if (joinEvent && leaveEvent) {
```
jivdhaliwal, line 43: Do we need to check for both join and leave visit events? Have we considered simply counting the unique session IDs grouped by visit ID and then getting the average count from there ?

chubberlisk: Nope, I didn't consider just using the session ID! I did consider grouping in the SQL statement like I said in the Notes but as a first pass I decided to write the simplest SQL statement first (as it's not something I'm very knowledgeable about so didn't want to get bogged down on perfecting a statement) and use JS to do the rest.

If doing things in Postgres is the way we want to go, could we 🍐 on it?

chubberlisk: I think it turned out easier than I thought! Just pushed up https://github.com/madetech/nhs-virtual-visit/pull/455/commits/e03f0f20c13b09170b8b6490dbb8b06947915f1a, let me know if this is what you were thinking.

chubberlisk:
Closing this in favour of https://github.com/madetech/nhs-virtual-visit/pull/456 which includes displaying the metric on the dashboard. Not much gain by splitting the feature into 2 PRs as it doesn't make it much bigger.

PR #456, opened 2020-06-24
Display average participants in a visit on trust admin dashboard

chubberlisk:
# What

Display average participants in a visit on trust admin dashboard by adding in a `retrieveAverageParticipantsInVisit` usecase.

# Why

So that a trust admin can see the average number of participants in a visit.

# Screenshots

| Before | After |
|--------|-------|
|image|image|

# Notes

N/A

### DIFF
--- a/pages/trust-admin.js
+++ b/pages/trust-admin.js
@@ -23,7 +23,9 @@ const TrustAdmin = ({
   visitsScheduled,
 }) => {
   if (wardError || trustError || averageParticipantsInVisitError) {
-    return <Error />;
+    return (
+      <Error err={wardError || trustError || averageParticipantsInVisitError} />
+    );
   }
 
   return (
--- a/pages/trust-admin.js
+++ b/pages/trust-admin.js
@@ -18,9 +18,11 @@ const TrustAdmin = ({
   leastVisited,
   mostVisited,
   trust,
+  averageParticipantsInVisit,
+  averageParticipantsInVisitError,
   visitsScheduled,
 }) => {
-  if (wardError || trustError) {
+  if (wardError || trustError || averageParticipantsInVisitError) {
     return <Error />;
   }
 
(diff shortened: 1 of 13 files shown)


pages/trust-admin.js
```diff
@@ -18,9 +18,11 @@ const TrustAdmin = ({
   leastVisited,
   mostVisited,
   trust,
+  averageParticipantsInVisit,
+  averageParticipantsInVisitError,
   visitsScheduled,
 }) => {
-  if (wardError || trustError) {
+  if (wardError || trustError || averageParticipantsInVisitError) {
     return <Error />;
```
jivdhaliwal, line 26: Is it time to consider grouping up the errors so that we can pass them through to the Error component? That way the error will be sent to Sentry and we will be notified

src/usecases/retrieveAverageParticipantsInVisit.js
```diff
@@ -0,0 +1,23 @@
+const retrieveAverageParticipantsInVisit = ({ getDb }) => async (trustId) => {
+  if (!trustId) return { error: "A trustId must be provided." };
+
+  const db = await getDb();
+  const [{ average_participants: averageParticipantsInVisit }] = await db.any(
+    `SELECT AVG(participants) as average_participants
```
jivdhaliwal, line 6: If there are 0 events in the database, I believe this query will return null as participants will be NULL and AVG removes all NULL results before calculating the average so it will also return NULL.

One option could be to use COALESCE to default the result of the AVG query to 0 if it returns NULL 
e.g. `SELECT COALESCE(AVG(participants), 0) as average_participants`

jivdhaliwal, line 6: What is returns if there is a scheduled visit with 3 participants and another visit with 0 (e.g. visit hasn't occurred yet or was cancelled)? Does it return 3 or 1.5 as the average?

jivdhaliwal, line 6: How precise do we want the result? 2 decimal places ? We can define the precision of the AVG result with https://www.postgresqltutorial.com/postgresql-avg-function/

Robware: Is the average to a decimal place even useful in this context?

chubberlisk: Good question! I'll write a test for this and update so it returns 3.

chubberlisk: Nice spot, updated and added a test for this as well.

pages/trust-admin/hospitals/[id].js
```diff
@@ -37,13 +37,13 @@ const ShowHospital = ({
           <Heading>{hospital.name}</Heading>
           <GridRow className="nhsuk-u-padding-bottom-3">
             <GridColumn
-              className="nhsuk-u-padding-bottom-3 nhsuk-u-one-half"
+              className="nhsuk-u-padding-bottom-3 nhsuk-grid-column-one-half"
```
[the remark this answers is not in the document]
Comment (chubberlisk) [reply pages/trust-admin/hospitals/[id].js:40]: Ah that makes sense! Will put it back to using `nhsuk-u-one-half`.

src/usecases/retrieveAverageParticipantsInVisit.js
chubberlisk: Just checked it returns 3! We don't have this problem here because there won't be any events for that visit so no events to group for that visit.

pages/trust-admin.js
chubberlisk: Not 100% sure what you mean but had a look at how it's used elsewhere. I think you mean passing down these errors into the `<Error err={error}` component. How do you imagine them being grouped? 🤔

jivdhaliwal: Yes sorry for being vague, I meant doing `<Error err={error}>`. It only really takes one exception object so since we have 3 we could pass them all in and let it handle returning the first one that is not null (`<Error err={wardError || trustError || averageParticipantsInVisitError}>`

chubberlisk: No worries! That makes sense, will update.

PR #459, opened 2020-06-25
Usecase for calculating the average visit time per trust

jivdhaliwal:
# What
Adds a usecase to calculate the average visit time per trust

# Why
So that we can show the average visit time metric to Trust Admins

# Screenshots

# Notes
We only calculate the call duration if a visit has both a join-visit and leave-visit event, all other visits are ignored


jivdhaliwal:
> Tested locally and returns what I'd expect! I also like the explicit tests for the edge cases, I couldn't think of anymore than you already have. 🎉
> 
> I have two thoughts and was wondering what you think.
> 
> * I've started to have mixed feelings about how I wrote `src/usecases/retrieveAverageParticipantsInVisit.contractTest.js` and reading the contract test you added. I think they're quite noisy and hard to read, like seeing what's the setup and the thing we're testing takes time as you have to read through the whole test. Wondering if it would be helpful to extract methods in the test (e.g. `visitWithAJoinAndLeaveEvents()` or something) or create more factories.

Ye I agree the setup is quite noisy. Factories will help a lot with this. I can add some in this PR to help with the noise and then we can open another PR to use those factories elsewhere.

> * I think working out metrics using SQL is cool but if one was any more bigger or complex than this one I'd begin to feel uncomfortable with the idea. Thinking about how easy to it is to understand, debugging, testing etc. in comparison to using JS and extracting well named methods.

Out of all the metrics we have this is probably the most complex so I get where you're coming from. I disagree on doing the same in Javascript being less complex though, as it can take a lot more code to do something similar in Javascript (for example there is no standard library way of finding the average value in an array, you have to write your own implementation)

chubberlisk:
>Ye I agree the setup is quite noisy. Factories will help a lot with this. I can add some in this PR to help with the noise and then we can open another PR to use those factories elsewhere.

Sounds perfect!

>Out of all the metrics we have this is probably the most complex so I get where you're coming from. I disagree on doing the same in Javascript being less complex though, as it can take a lot more code to do something similar in Javascript (for example there is no standard library way of finding the average value in an array, you have to write your own implementation). But if we do pass a threshold of complexity where we are no longer comfortable using a SQL query then I agree we should re-evaluate our implementation and focus on simplicity and ease of understanding.

Nice, just wanted to raise it as a thing in my brain. :)

PR #462, opened 2020-06-29
Add reporting for booked visits start date to trust admin dashboard

chubberlisk:
# What

Add reporting for booked visits start date to trust admin dashboard.

# Why

We received feedback that "Without a date reference, the number of bookings isn’t particularly helpful." so by adding the start date they'll be able to know how many scheduled visits since a given date.

# Screenshots

Without any booked visits | With booked visits
-- | --
image|image

# Notes

Would like feedback on the last two commits which adds a new factory call `setupWardWithinHospitalAndTrust` and uses it in the contract test for the new usecase.

### DIFF
--- a/src/testUtils/factories.js
+++ b/src/testUtils/factories.js
@@ -26,6 +26,18 @@ export const setupWard = async (args = {}) => {
   });
 };
 
+export const setupWardWithinHospitalAndTrust = async ({ index = 1 }) => {
+  const { trustId } = await setupTrust({ adminCode: `TESTCODE${index}` });
+  const { hospitalId } = await setupHospital({ trustId });
+  const { wardId } = await setupWard({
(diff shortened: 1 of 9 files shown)


src/testUtils/factories.js
```diff
@@ -26,6 +26,18 @@ export const setupWard = async (args = {}) => {
   });
 };
 
+export const setupWardWithinHospitalAndTrust = async ({ index = 1 }) => {
```
jivdhaliwal, line 29: Great way to tackle the unique constraints 👍

chubberlisk:
> New factories \o/
> 
> What do we think about placing the timestamp at the bottom of the dashboard instead of the top ? Feels a bit distracting up there

Yup, I get what you mean. Will move downstairs.

PR #463, opened 2020-06-29
Show average visit time metric

jivdhaliwal:
# What
Shows the average visit time metric on the trust admin dashboard

# Why
So that trust admins can get an overview of their trusts performance

# Screenshots
ScreenShot 2020-06-29 at 17 10 49

# Notes

### DIFF
--- a/pages/trust-admin.js
+++ b/pages/trust-admin.js
@@ -200,7 +200,7 @@ export const getServerSideProps = propsWithContainer(
 
     const averageVisitTime = moment
       .duration(averageVisitTimeSeconds, "seconds")
-      .format("h [hours], m [minutes]");
+      .format("h[hr], m[min]");
 
     const error =
       wardError ||
--- a/pages/trust-admin.js
+++ b/pages/trust-admin.js
@@ -13,30 +13,18 @@ import { TRUST_ADMIN } from "../src/helpers/userTypes";
 import formatDate from "../src/helpers/formatDate";
 
 const TrustAdmin = ({
-  wardError,
-  trustError,
+  error,
   wards,
   hospitals,
   leastVisited,
   mostVisited,
   trust,
   averageParticipantsInVisit,
-  averageParticipantsInVisitError,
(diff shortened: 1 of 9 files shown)


pages/trust-admin.js
```diff
@@ -163,24 +175,41 @@ export const getServerSideProps = propsWithContainer(
...
 
+    const error =
+      wardError ||
+      trustError ||
+      averageParticipantsInVisitError ||
+      averageVisitTimeSecondsError;
+
+    const averageVisitTime = moment
+      .duration(averageVisitTimeSeconds, "seconds")
+      .format("h[hr], m[min]");
```
chubberlisk, line 200: What do we think about adding a space between the number and `hr`/`min`?

image

pages/trust-admin.js
```diff
@@ -9,23 +9,23 @@ import NumberTile from "../src/components/NumberTile";
 import Text from "../src/components/Text";
 import AnchorLink from "../src/components/AnchorLink";
 import { TRUST_ADMIN } from "../src/helpers/userTypes";
+import moment from "moment";
+import momentDurationFormatSetup from "moment-duration-format";
+momentDurationFormatSetup(moment);
```
chubberlisk, line 14: Non-blocking: We could wrap this in a helper like we've done with formatting date and time.

pages/trust-admin.js
jivdhaliwal: Good shout that reads much better

pages/trust-admin.js
jivdhaliwal: Ye I agree it would be good to get the duration calculation out of the page. Since it's only being used in the usecase, do we think that would be a suitable place to put it for now ? (also because it's formatted specifically for this usecase so that the string fits within the number tile)

chubberlisk: Works for me! We can always move it out if it's needed elsewhere.

PR #465, opened 2020-06-29
Average daily visits usecase

Robware:
# What
Adds a usecase for calculating the average daily visits
# Why
To retrieve the average daily visits for a trust on the metrics page
# Screenshots

# Notes
This includes the jest config changes to ignore the docker-data directory because I couldn't do a PR earlier.

### DIFF
--- a/src/usecases/retrieveAverageVisitsPerDay.contractTest.js
+++ b/src/usecases/retrieveAverageVisitsPerDay.contractTest.js
@@ -1,5 +1,5 @@
 import AppContainer from "../containers/AppContainer";
-import { setupTrust } from "../testUtils/factories";
+import { setupTrust, setupHospital, setupWard } from "../testUtils/factories";
 import { v4 as uuidv4 } from "uuid";
 
 describe("retrieveAverageVisitsPerDay contract tests", () => {
@@ -9,12 +9,12 @@ describe("retrieveAverageVisitsPerDay contract tests", () => {
     // A trust with a visit with 1 participant
     const { trustId } = await setupTrust();
 
-    const { hospitalId } = await container.getCreateHospital()({
+    const { hospitalId } = await setupHospital({
       name: "Test Hospital",
       trustId: trustId,
     });
 
-    const { wardId } = await container.getCreateWard()({
+    const { wardId } = await setupWard({
       name: "Test Ward 1",
       code: "wardCode1",
       hospitalId: hospitalId,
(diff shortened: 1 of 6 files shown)


src/usecases/retrieveAverageVisitsPerDay.js
```diff
@@ -0,0 +1,24 @@
...
+  const db = await getDb();
+  const [{ average_visits_per_day: averageVisitsPerDay }] = await db.any(
+    `SELECT to_char(COALESCE(AVG(visits), 0), '999D9') as average_visits_per_day FROM(
+      SELECT COUNT (distinct events.visit_id ) AS visits
+      FROM events
+      LEFT JOIN scheduled_calls_table ON events.visit_id = scheduled_calls_table.id
+      LEFT JOIN wards ON scheduled_calls_table.ward_id = wards.id
+      LEFT JOIN trusts ON wards.trust_id = trusts.id
+      WHERE trusts.id = $1 AND events.action='join-visit'
+      GROUP BY date(scheduled_calls_table.call_time)
```
jivdhaliwal, line 13: Should we be grouping by the call_time or the time of the event ?

src/usecases/retrieveAverageVisitsPerDay.contractTest.js
```diff
@@ -0,0 +1,179 @@
+import AppContainer from "../containers/AppContainer";
+import { setupTrust } from "../testUtils/factories";
+import { v4 as uuidv4 } from "uuid";
+
+describe("retrieveAverageVisitsPerDay contract tests", () => {
+  const container = AppContainer.getInstance();
+
+  it("returns the average number of visits per day for a trust", async () => {
```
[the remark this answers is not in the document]
Robware: This is not available to me until I pull in `master`. Once I get up to date I'll take a look at it.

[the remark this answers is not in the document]
Robware: Updated to use appropriate factories

src/usecases/retrieveAverageVisitsPerDay.js
Robware: Grouping by `call_time`. Event time doesn't work.

jivdhaliwal: Could we clarify why it doesn't work ? Is it not suitable for this metric?

Robware: I've not got to the bottom of why, but for some reason the values come back all wrong.

jivdhaliwal: Were we using MockDate to set the correct date before capturing events ? (like https://github.com/madetech/nhs-virtual-visit/blob/master/src/usecases/retrieveAverageVisitTimeByTrustId.contractTest.js#L40-L54)

Robware: So I took a look at this and the test pass with the mock date (woo). However running the query against the staging data set still gives wildly different results. Not sure what route to do down with this as it needs investigation to determine which is best, but I'm not sure it's worth the time. What were you expecting to gain from using event time vs call time?

jivdhaliwal: Just wanted to make sure we were using the right timestamp for the metric (as we use the event time for other metrics). But since this one is for daily visits it makes sense to use the call_time as it's unlikely we'll have people joining days after the call is scheduled (since the data is wiped)

PR #469, opened 2020-06-30
Add migration to alter time in events to be timestamp with time zone

chubberlisk:
# What

Alter the `time` column in the `events` table to be of type `timestamp with time zone` instead of `timestamp without time zone`.

# Why

This was causing issues when retrieving the time from the events table.

# Screenshots

N/A

# Notes

N/A


db/migrations/sqls/20200630113809-alter-time-type-in-events-up.sql
```diff
@@ -0,0 +1,4 @@
+/* Replace with your SQL commands */
+
+ALTER TABLE events
+ALTER time type timestamptz USING time AT TIME ZONE 'UTC'
```
jivdhaliwal, line 4: The up migration is converting UTC times to the current time zone.

i.e.. "2020-06-30 12:00:00.000" is changing to "2020-06-30 13:00:00.000+01"

The down migration is then stripping the timezone but maintaining the same time

i.e.  "2020-06-30 13:00:00.000+01" changes to "2020-06-30 13:00:00.000"

That means that if you keep running the migrations up and down, the time will keep creeping up instead of staying the same

jivdhaliwal: I think we either need to ensure the up migration maintains a UTC offset (instead of BST / local timezone) or that the down migration converts back to UTC before stripping the offset. My vote would be to ensure the up maintains a UTC offset

PR #472, opened 2020-06-30
Add other reporting start date on the trust admin dashboard

chubberlisk:
# What

Add reporting start date on the trust admin dashboard.

# Why

To provide context for when reporting for metrics such as average visit time. average participants, etc. started.

# Screenshots

Before | After
-- | --
image|image

# Notes

N/A

### DIFF
--- a/pages/trust-admin.js
+++ b/pages/trust-admin.js
@@ -146,15 +146,17 @@ const TrustAdmin = ({
               </div>
             </GridColumn>
           </GridRow>
-          {(wardVisitTotalsStartDate || reportingStartDate) && (
-            <ReviewDate>
-              {wardVisitTotalsStartDate &&
-                `Start date for booked visits reporting: ${wardVisitTotalsStartDate}`}
-              {wardVisitTotalsStartDate && reportingStartDate && <br />}
-              {reportingStartDate &&
-                `Start date for other reporting: ${reportingStartDate}`}
-            </ReviewDate>
-          )}
+
+          <ReviewDate
+            className="nhsuk-u-margin-bottom-0"
+            beforeDateText="Start date for booked visits reporting: "
+            date={wardVisitTotalsStartDate}
+          />
+          <ReviewDate
+            className="nhsuk-u-margin-0"
+            beforeDateText="Start date for other reporting: "
+            date={reportingStartDate}
+          />
         </GridColumn>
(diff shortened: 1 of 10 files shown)


pages/trust-admin.js
```diff
@@ -145,9 +146,13 @@ const TrustAdmin = ({
...
           </GridRow>
-          {wardVisitTotalsStartDate && (
+          {(wardVisitTotalsStartDate || reportingStartDate) && (
             <ReviewDate>
-              Reporting for booked visits start date: {wardVisitTotalsStartDate}
+              {wardVisitTotalsStartDate &&
+                `Start date for booked visits reporting: ${wardVisitTotalsStartDate}`}
+              {wardVisitTotalsStartDate && reportingStartDate && <br />}
+              {reportingStartDate &&
+                `Start date for other reporting: ${reportingStartDate}`}
```
chubberlisk, line 155: A bit grim with all the conditionals. Any thoughts on refactoring?

jivdhaliwal: ReviewDate

jivdhaliwal: What if we used a ReviewDate component for each date and let the ReviewDate component handle the case where the date is null

pages/trust-admin.js
```diff
@@ -182,11 +187,18 @@ export const getServerSideProps = propsWithContainer(
...
+    const retrieveReportingStartDateResponse = await container.getRetrieveReportingStartDateByTrustId()(
+      authenticationToken.trustId
+    );
 
     const wardVisitTotalsStartDate = retrieveWardVisitTotalsStartDateResponse.startDate
       ? formatDate(retrieveWardVisitTotalsStartDateResponse.startDate)
       : null;
 
+    const reportingStartDate = retrieveReportingStartDateResponse.startDate
+      ? formatDate(retrieveReportingStartDateResponse.startDate)
```
jivdhaliwal, line 199: Would it make sense to format the date inside the usecase instead of the page? Or do we want to keep that logic separate?

chubberlisk: When doing the original booking visits start date, my thoughts were formatting the date was a presentational thing so kept it on the page. But looking now and after we moved the moment duration code to the average visit time usecase, it just feels like noise on this page that could be moved to the usecase...

PR #482, opened 2020-07-02
Show support link at the end visits page for a key contact

chubberlisk:
# What

Show support link at the end visits page for a key contact.

# Why

So that a key contact can get support if they need it.

# Screenshots

Before | After
-- | --
image|screencapture-localhost-3000-visits-end-2020-07-02-11_55_57

# Notes

Have dropped a number of screenshots for different content and design variations I tried on Slack.

### DIFF
--- a/pages/visits/end.js
+++ b/pages/visits/end.js
@@ -1,5 +1,6 @@
 import React from "react";
 import Layout from "../../src/components/Layout";
+import ActionLinkSection from "../../src/components/ActionLinkSection";
 import ActionLink from "../../src/components/ActionLink";
 import AnchorLink from "../../src/components/AnchorLink";
 import InsetText from "../../src/components/InsetText";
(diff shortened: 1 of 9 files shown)


pages/visits/end.js
```diff
@@ -50,10 +50,22 @@ const End = ({ wardId, callId, surveyUrl }) => {
...
+            <h2>What happens next</h2>
+
+            <p>Your personal data will be removed within 24 hours.</p>
+
+            {supportUrl && (
+              <ActionLink href={supportUrl}>
+                Get support from this hospital
+              </ActionLink>
+            )}
+
```
Robware, line 68: Do you think it's worth wrapping this, and the similar code below, up in a component?

chubberlisk: Are you thinking like a `WardVisitEnd` component for the ward staff and then `VisitorVisitEnd` component for the visitor/key contact?

Robware: Sorry, I failed to select the full section of code for the comment, which would've made it clearer.

I'm talking that having an `ExtraInformation` component, or something, would be worthwhile as the _What happens Next_ and _Help improve virtual visits_ sections are the same. Having them in a component would mean that they stay consistent and all you'd need to do is pass in the text to populate it.

chubberlisk: Ah okay! Something like the below?

```jsx
<ActionLinkSection heading="What happens next" link={supportUrl} />

<ActionLinkSection heading="Help improve virtual visits" link={surveyUrl}>
  <p>
    We’d welcome your feedback. Can you answer some questions
    about your virtual visit today?
  </p>
</ActionLinkSection>
```

chubberlisk: Pushed up two new commits with this, let me know what you think!

Robware: Late to the party, but that's perfect!

PR #483, opened 2020-07-02
Add survey URL to the edit hospital form

jivdhaliwal:
# What

Adds a  survey URL field to the edit hospital form

# Why

So that TrustAdmins can add a way for key contacts to give feedback at the end of a call

# Screenshots

ScreenShot 2020-07-02 at 11 03 49

# Notes

### DIFF
--- a/src/components/EditHospitalForm/index.js
+++ b/src/components/EditHospitalForm/index.js
@@ -91,7 +91,7 @@ const EditHospitalForm = ({ errors, setErrors, hospital }) => {
             type="text"
             hasError={hasError("hospital-name")}
             errorMessage={errorMessage("hospital-name")}
-            className="nhsuk-u-font-size-32 nhsuk-input--width-10"
+            className="nhsuk-u-font-size-32"
             style={{ padding: "16px!important", height: "64px" }}
             onChange={(event) => setHospitalName(event.target.value)}
             name="hospital-name"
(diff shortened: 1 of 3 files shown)


src/components/EditHospitalForm/index.js
```diff
@@ -96,6 +100,28 @@ const EditHospitalForm = ({ errors, setErrors, hospital }) => {
           />
         </FormGroup>
 
+        <FormGroup>
+          <Label htmlFor="hospital-survey-url" className="nhsuk-label--l">
+            Key contact survey URL
```
chubberlisk, line 105: Could we add `(optional)` to make it clear to the user that this isn't required?

```suggestion
            Key contact survey URL (optional)
```

This is the suggestion way by the NHS Design System. See https://service-manual.nhs.uk/content/how-to-write-good-questions-for-forms/make-sure-you-need-each-question#start-with-essential-questions.

src/components/EditHospitalForm/index.js
```diff
@@ -96,6 +100,28 @@ const EditHospitalForm = ({ errors, setErrors, hospital }) => {
...
+          <Input
+            id="hospital-survey-url"
+            type="url"
+            hasError={hasError("hospital-survey-url")}
+            errorMessage={errorMessage("hospital-survey-url")}
+            className="nhsuk-u-font-size-32"
+            style={{ padding: "16px!important", height: "64px" }}
+            onChange={(event) => setHospitalSurveyUrl(event.target.value)}
+            name="hospital-survey-url"
+            autoComplete="off"
```
chubberlisk, line 120: It might be useful to have autocomplete on for this?

src/components/EditHospitalForm/index.js
jivdhaliwal: Ah thanks I was wondering how to display the optionality 🙌

src/components/EditHospitalForm/index.js
jivdhaliwal: Eagle eye! Good catch

jivdhaliwal:
> Just noticed we don't display the survey link anywhere and two smol comments below.

I've updated the main card with a task to display the URL in the trust admin dashboard since there are a few places

PR #484, opened 2020-07-02
Create helper to validate URLs

Robware:
# What
Create helper to validate URLs
# Why
So we can make sure that the user only enters a valid URL
# Screenshots

# Notes

### DIFF
--- a/src/helpers/validateUrl.test.js
+++ b/src/helpers/validateUrl.test.js
@@ -1,0 +1,95 @@
+import validateUrl from "./validateUrl";
+
+const validUrls = [
+  "http://foo.com/blah_blah",
+  "http://foo.com/blah_blah/",
+  "http://foo.com/blah_blah_(wikipedia)",
+  "http://foo.com/blah_blah_(wikipedia)_(again)",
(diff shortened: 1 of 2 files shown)


src/helpers/validateUrl.test.js
```diff
@@ -0,0 +1,95 @@
+import validateUrl from "./validateUrl";
+
+const validUrls = [
+  "http://foo.com/blah_blah",
+  "http://foo.com/blah_blah/",
+  "http://foo.com/blah_blah_(wikipedia)",
+  "http://foo.com/blah_blah_(wikipedia)_(again)",
+  "http://www.example.com/wpstyle/?p=364",
+  "https://www.example.com/foo/?bar=baz&inga=42&quux",
```
chubberlisk, line 9: Could we move the `https` example to the top of the array or add more `https` examples? To make it clearer that it's supported in the test. I was about to comment that we haven't got any examples of it until I found one buried in the sea of `http`! 😅

Robware: Done :)

PR #485, opened 2020-07-02
Show survey url in hospitals table

jivdhaliwal:
# What
Shows the survey URL for a hospital in the hospitals table if one is set

# Why
So that the TrustAdmin can easily review the survey URLs set on their hospitals

# Screenshots
ScreenShot 2020-07-02 at 14 58 55

# Notes

### DIFF
--- a/src/components/HospitalsTable/index.js
+++ b/src/components/HospitalsTable/index.js
@@ -17,6 +17,9 @@ const HospitalsTable = ({ hospitals }) => (
             Booked visits
           </th>
           <th className="nhsuk-table__header" scope="col">
+            Survey URL
+          </th>
+          <th className="nhsuk-table__header" scope="col">
             <span className="nhsuk-u-visually-hidden">Actions</span>
           </th>
         </tr>
@@ -27,6 +30,13 @@ const HospitalsTable = ({ hospitals }) => (
             <td className="nhsuk-table__cell">{hospital.name}</td>
             <td className="nhsuk-table__cell">{hospital.wards.length}</td>
             <td className="nhsuk-table__cell">{hospital.bookedVisits}</td>
+            <td className="nhsuk-table__cell">
+              {hospital.surveyUrl ? (
+                <a href={hospital.surveyUrl}>Link</a>
+              ) : (
+                "None"
+              )}
+            </td>
             <td className="nhsuk-table__cell" style={{ textAlign: "center" }}>
               <AnchorLink
                 href={`/trust-admin/hospitals/${hospital.id}`}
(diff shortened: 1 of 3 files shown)


src/components/HospitalsTable/index.js
```diff
@@ -27,6 +30,13 @@ const HospitalsTable = ({ hospitals }) => (
             <td className="nhsuk-table__cell">{hospital.name}</td>
             <td className="nhsuk-table__cell">{hospital.wards.length}</td>
             <td className="nhsuk-table__cell">{hospital.bookedVisits}</td>
+            <td className="nhsuk-table__cell">
+              {hospital.surveyUrl ? (
+                <a href={hospital.surveyUrl}>Link</a>
```
chubberlisk, line 35: As a user, I'd prefer to see the value of the link instead of having to click `Link` to find out what link is e.g. is it the Google Form we set up or the Typeform 🤔 .

However, I realise this raises the issue of long links in a table column (also bearing in mind that we want the show the support link too). Two ideas I could think of to alleviate are:

- Remove the number of wards column (might help not sure)
- Only display the support and survey links on an individual hospital page.

A super quick browser HTML hackaroo:

image

The Change links will link to the `Edit hospital` page which we haven't got a link for on the hospital page yet, only on the list of hospitals page.

What do you think?

jivdhaliwal: I like showing more info on the hospital page, but even in that case the link will be too long to display (one of the current survey links is 130 characters long). In that case we could trim the content to a fixed width and leave an ellipsis at the end (and possible reveal the full link on mouse over). 

If we do that we'll still end up with links that all look the same (e.g. "https://forms.office.com/Pages/..."), so would we also want the url to appear when you mouse over?

chubberlisk: Hmm, that's a good point. I'd like to avoid showing content when mouse over as I worry about the accessibility of it and I'm not sure if that'll work on mobile devices.

How about we stick with this for now but with one smol improvement for screen readers.

```suggestion
                <a href={hospital.surveyUrl}>Link<span className="nhsuk-u-visually-hidden"> for {hospital.name} survey</span></a>
```

jivdhaliwal: Ye the mouse over doesn't feel very accessible, not a fan of the hidden functionality. Based on the survey links we have so far, I feel like there's little to gain from knowing the contents just yet as they are all pretty much identical so I don't think we're losing out much on just showing the word "Link"

I like the screen reader recommendation 👍  will make that change now

PR #488, opened 2020-07-03
Add display of survey and support URL on hospital page

chubberlisk:
# What

Add display of survey and support URL on hospital page.

# Why

Currently we don't show all information of a hospital on their individual pages.

# Screenshots

Before | After
-- | --
image|image

# Notes

N/A

### DIFF
--- a/pages/trust-admin/hospitals/[id].js
+++ b/pages/trust-admin/hospitals/[id].js
@@ -10,6 +10,7 @@ import Layout from "../../../src/components/Layout";
 import WardsTable from "../../../src/components/WardsTable";
 import NumberTile from "../../../src/components/NumberTile";
 import Panel from "../../../src/components/Panel";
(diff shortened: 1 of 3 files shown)


Robware:
An I right in thinking that `Change` link doesn't do anything? If so, is there a way to hide it?

chubberlisk:
> Am I right in thinking that `Change` link doesn't do anything? If so, is there a way to hide it?

The `Change` link takes you to the edit hospital form when shown on the hospital page. Yush, the component can hide the `Change` links if `withActions={false}` but in this case it's useful to have them.

Robware:
> > Am I right in thinking that `Change` link doesn't do anything? If so, is there a way to hide it?
> 
> The `Change` link takes you to the edit hospital form when shown on the hospital page. Yush, the component can hide the `Change` links if `withActions={false}` but in this case it's useful to have them.

Ah, yes, sorry. I read the default value of the property, and not the valued passed in.

PR #496, opened 2020-07-06
Move the duplicated isPresent function to its own helper

Robware:
# What
Move the duplicated isPresent function to its own helper
# Why
Duplicate code is bad
# Screenshots

# Notes

### DIFF
--- a/src/helpers/isPresent.js
+++ b/src/helpers/isPresent.js
@@ -1,0 +1,5 @@
+export default (input) => {
+  if (input && input.length !== 0) {
+    return input;
+  }
+};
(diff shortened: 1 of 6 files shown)


src/helpers/isPresent.js
```diff
@@ -0,0 +1,5 @@
+export default (input) => {
```
chubberlisk, line 1: What do you think about adding a quick and simple unit test for this?

Robware: Development driven tests. It did cross my mind, yes. I'll pop one in.

PR #505, opened 2020-07-08
Edit hospital form refactor

Robware:
# What
Refactor the hospital edit and add code to both use the `EditHospitalForm`
# Why
Reduce duplication and thus maintenance and fragility
# Screenshots

# Notes

### DIFF
--- a/pages/trust-admin/hospitals/add.js
+++ b/pages/trust-admin/hospitals/add.js
@@ -71,7 +71,6 @@ const AddAHospital = ({ error, trustId }) => {
           <EditHospitalForm
             errors={errors}
             setErrors={setErrors}
-            hospital={{}}
             submit={submit}
           />
         </GridColumn>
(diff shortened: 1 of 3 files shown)


pages/trust-admin/hospitals/add.js
```diff
@@ -111,49 +68,12 @@ const AddAHospital = ({ error, trustId }) => {
...
-            </FormGroup>
-
-            <Button className="nhsuk-u-margin-top-5">Add hospital</Button>
-          </form>
+          <EditHospitalForm
+            errors={errors}
+            setErrors={setErrors}
+            hospital={{}}
+            submit={submit}
+          />
```
chubberlisk, line 76: 🎉

chubberlisk: Nitpick: I think if we update `EditHospitalForm` to `const EditHospitalForm = ({ errors, setErrors, hospital = {}, submit }) => {` with a default value, it'll meet that we won't have to do `hospital={{}}`.

Robware: Good spot, will do that.

PR #512, opened 2020-07-13
Send visit updated notifications via the sendBookingNotification usecase

jivdhaliwal:
# What
Updates the sendBookingNotification usecase to be able to send 'visit updated' notifications as well as 'new visit' notifications.

# Why
Allows us to use this usecase in the upcoming update-a-visit API to send notifications

# Screenshots

# Notes

### DIFF
--- a/src/usecases/sendBookingNotification.js
+++ b/src/usecases/sendBookingNotification.js
@@ -14,8 +14,8 @@ const sendBookingNotification = ({
   hospitalName,
   visitDateAndTime,
 }) => {
-  const textMessageTemplateId = TemplateStore.firstText.templateId;
-  const emailTemplateId = TemplateStore.firstEmail.templateId;
+  const textMessageTemplateId = TemplateStore().firstText.templateId;
+  const emailTemplateId = TemplateStore().firstEmail.templateId;
 
   const personalisation = {
     visit_date: formatDate(visitDateAndTime),
--- a/src/usecases/sendBookingNotification.js
+++ b/src/usecases/sendBookingNotification.js
@@ -5,7 +5,8 @@ import formatTime from "../../src/helpers/formatTime";
 import ConsoleNotifyProvider from "../providers/ConsoleNotifyProvider";
 
 const NEW_NOTIFICATION = "new";
-const NOTIFICATION_TYPES = [NEW_NOTIFICATION];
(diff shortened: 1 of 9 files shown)


src/usecases/sendBookingNotification.js
```diff
@@ -69,4 +78,21 @@ const sendBookingNotification = ({
...
+
+  if (notificationType == NEW_NOTIFICATION) {
+    textMessageTemplateId = TemplateStore().firstText.templateId;
+    emailTemplateId = TemplateStore().firstEmail.templateId;
+  }
+
+  if (notificationType == UPDATED_NOTIFICATION) {
+    textMessageTemplateId = TemplateStore().updatedVisitText.templateId;
+    emailTemplateId = TemplateStore().updatedVisitEmail.templateId;
+  }
```
Robware, line 93: How about converting this to a switch statement, then throwing an exception on the default case for when it's passed something not recognised?

src/usecases/sendBookingNotification.test.js
```diff
@@ -29,6 +32,27 @@ describe("sendBookingNotification", () => {
...
+    emailUpdatedTemplateId = TemplateStore().updatedVisitEmail.templateId;
+  });
+
+  afterEach(() => {
+    process.env.SMS_INITIAL_TEMPLATE_ID = "";
+    process.env.SMS_UPDATED_VISIT_TEMPLATE_ID = "";
+    process.env.SMS_JOIN_TEMPLATE_ID = "";
+    process.env.EMAIL_INITIAL_TEMPLATE_ID = "";
+    process.env.EMAIL_UPDATED_VISIT_TEMPLATE_ID = "";
+    process.env.EMAIL_JOIN_TEMPLATE_ID = "";
```
Robware, line 55: If the `TemplateStore` is being injected now, would it not be possible to mock it out so we don't need to reset the env vars between each test?

jivdhaliwal: I decided not to go down the route of injecting the template store via the app container in the end and went for simpler implementation (just changing it to a function and setting env variables in the test)

src/usecases/sendBookingNotification.js
jivdhaliwal: Yep good idea, that keeps all the notification type logic in one place

PR #517, opened 2020-07-16
Update a visit API

Robware:
# What
API backend to update a visit
# Why
So that users can edit visit details
# Screenshots

# Notes
Work still pending on sending notifications

### DIFF
--- a/pages/api/update-a-visit.js
+++ b/pages/api/update-a-visit.js
@@ -1,0 +1,69 @@
+import withContainer from "../../src/middleware/withContainer";
+import validateVisit from "../../src/helpers/validateVisit";
+
+export default withContainer(
+  async ({ headers, body, method }, res, { container }) => {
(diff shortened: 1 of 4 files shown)


pages/api/update-a-visit.js
```diff
@@ -0,0 +1,69 @@
...
+      contactName: body.contactName,
+      contactEmail: body.contactEmail,
+      contactNumber: body.contactNumber,
+      callTime: body.callTime,
+    };
+
+    try {
+      await container.getUpdateVisitByCallId()(updatedCall);
+      respond(200, { success: true });
+    } catch (updateError) {
```
jivdhaliwal, line 64: Do we need to send these errors to sentry / console.error ?

Robware: Good spot, will add that now

PR #519, opened 2020-07-16
Update a visit notifications

Robware:
# What
Send notifications when updating a visit
# Why
So that the relevant people are informed when things change
# Screenshots

# Notes

### DIFF
--- a/pages/api/update-a-visit.js
+++ b/pages/api/update-a-visit.js
@@ -1,5 +1,29 @@
 import withContainer from "../../src/middleware/withContainer";
 import validateVisit from "../../src/helpers/validateVisit";
+import {
+  NEW_NOTIFICATION,
+  UPDATED_NOTIFICATION,
+} from "../../src/usecases/sendBookingNotification";
+
+const determineNotificationType = (sideA, sideB) => {
+  if (
+    sideA.callId == sideB.callId &&
+    sideA.patientName == sideB.patientName &&
+    sideA.recipientName == sideB.recipientName &&
+    sideA.recipientEmail == sideB.recipientEmail &&
+    sideA.recipientNumber == sideB.recipientNumber &&
+    sideA.callTime == sideB.callTime
+  )
+    return false;
+
+  if (
+    sideA.recipientEmail !== sideB.recipientEmail ||
+    sideA.recipientNumber !== sideB.recipientNumber
+  )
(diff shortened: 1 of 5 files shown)


pages/api/update-a-visit.js
```diff
@@ -1,5 +1,28 @@
 import withContainer from "../../src/middleware/withContainer";
 import validateVisit from "../../src/helpers/validateVisit";
+import {
+  NEW_NOTIFICATION,
+  UPDATED_NOTIFICATION,
+} from "../../src/usecases/sendBookingNotification";
+
+const determineNotificationType = (sideA, sideB) => {
```
chubberlisk, line 8: What happens if the original visit used an email address and the updated visit uses a mobile number (and vice versa)? I.e. Key contact is notified via email -> Key contact is notified via text message

What happens if the original visit used an email address and the updated visit adds a mobile number (and vice versa)? I.e. Key contact is notified via email -> Key contact is notified via email and text message

Robware: Currently, if either the email or phone number change then the new visit notification is sent out on all channels.

Thinking about it, this is less than idea, but not sure what the best approach to take is

jivdhaliwal: +1 for only sending the notification to the one that changed (otherwise it may appear like a new visit was booked to whoever already received the notification)

But we can tackle that in a separate PR to keep this one small

chubberlisk:
Had a thought whilst working on the success page for editing a visit - I think it would be useful to tell the ward staff whether or not a booking confirmation or updated visit notification has been sent. We can do this by returning the notification type as part of the API response. Of course not to be addressed in this PR but what do you think?

jivdhaliwal:
> Had a thought whilst working on the success page for editing a visit - I think it would be useful to tell the ward staff whether or not a booking confirmation or updated visit notification has been sent. We can do this by returning the notification type as part of the API response. Of course not to be addressed in this PR but what do you think?

I'm not sure how that would be of use to the Ward Staff? Do they need to know what kind of notification was sent?

Robware:
I do wonder what this would achieve over a simple message to say we've told contacts of the change. In any case, I don't think it's beyond the realms of possibility. We'd have to think about how we want to display sending out both a new and update notification should we do the change suggested on `update-a-visit.js`.

PR #525, opened 2020-07-20
Add gov.uk notifiy templates to the runbook

jivdhaliwal:
# What
Adds the notification template copy to the runbook

# Why
So that someone can set up the app with the correct templates

# Screenshots

# Notes

### DIFF
--- a/docs/runbook/govnotify-templates.md
+++ b/docs/runbook/govnotify-templates.md
@@ -1,0 +1,92 @@
+# Gov.uk Notify Templates
+
+https://www.notifications.service.gov.uk/using-notify/get-started
+
+The app uses 6 different notification templates (3 SMS and 3 Email). You can find suggested copy for each notification below, listed by the environment variable we use in the app.
+
+We suggest creating separate templates for staging and production so that you can test template changes without affecting the live app.
+
+## Templates
+
+### SMS_INITIAL_TEMPLATE_ID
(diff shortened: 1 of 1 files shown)


docs/runbook/govnotify-templates.md
```diff
@@ -0,0 +1,90 @@
+# Gov.uk Notify Templates
+
+The app uses 6 different notification templates (3 SMS and 3 Email). You can find suggested copy for each notification below, listed by the environment variable we use in the app.
+
+We suggest creating seperate templates for staging and production so that you can test template changes without affecting the live app.
```
chubberlisk, line 5:
```suggestion
We suggest creating separate templates for staging and production so that you can test template changes without affecting the live app.
```

docs/runbook/govnotify-templates.md
```diff
@@ -0,0 +1,90 @@
+# Gov.uk Notify Templates
```
chubberlisk, line 1: Would it be worth linking to the Notify docs? Like https://www.notifications.service.gov.uk/using-notify/guidance maybe?

docs/runbook/govnotify-templates.md
jivdhaliwal: Ah good catch. That was copied from prod so there's a typo in there

PR #529, opened 2020-07-21
Add edit visit page

chubberlisk:
# What

Add page for editing a visit.

# Why

So that a ward staff can update the details of a visit.

# Screenshots

image

# Notes

Next step to create the confirmation / check your answers page.

### DIFF
--- a/pages/wards/visits/[id]/edit.js
+++ b/pages/wards/visits/[id]/edit.js
@@ -1,0 +1,83 @@
+import React, { useState } from "react";
+import { GridRow, GridColumn } from "../../../../src/components/Grid";
+import Layout from "../../../../src/components/Layout";
+import ErrorSummary from "../../../../src/components/ErrorSummary";
+import VisitForm from "../../../../src/components/VisitForm";
+import verifyToken from "../../../../src/usecases/verifyToken";
+import Router from "next/router";
+import propsWithContainer from "../../../../src/middleware/propsWithContainer";
+import { WARD_STAFF } from "../../../../src/helpers/userTypes";
+
+const EditAVisit = ({
+  callId,
+  initialPatientName,
+  initialContactName,
+  initialContactNumber,
+  initialContactEmail,
+  initialCallDateTime,
+}) => {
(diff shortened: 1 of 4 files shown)


pages/wards/visits/[id]/edit.js
```diff
@@ -0,0 +1,83 @@
...
+  initialContactName,
+  initialContactNumber,
+  initialContactEmail,
+  initialCallDateTime,
+}) => {
+  const [errors, setErrors] = useState([]);
+
+  const submit = () => {
+    Router.push({
+      pathname: `/wards/visits/${callId}/edit-confirmation`,
```
Comment (jivdhaliwal) [inline pages/wards/visits/[id]/edit.js:23]: Since this page doesn't exist yet, should we redirect to the success page for now?

pages/wards/visits/[id]/edit.js
```diff
@@ -0,0 +1,83 @@
...
+          />
+        </GridColumn>
+      </GridRow>
+    </Layout>
+  );
+};
+export const getServerSideProps = propsWithContainer(
+  verifyToken(async ({ query, container }) => {
+    const callId = query.id;
+    const retrieveVisitByCallId = container.getRetrieveVisitByCallId();
```
Comment (jivdhaliwal) [inline pages/wards/visits/[id]/edit.js:56]: The `getRetrieveVisitByCallId` usecase is used in the patient / key contact visits pages to obfuscate the visit ID from the user. 

But here we're logged in as a ward staff so I feel the `[id]` in the path here should relate to the primary ID of the visit instead of the callID. What do you think?

Comment (chubberlisk) [reply pages/wards/visits/[id]/edit.js:56]: Makes sense!

Looks to me we'd have to do a couple of things if we want to do this including adding a new usecase to retrieve a visit by the `id` and maybe change our update a visit usecase to do it by `id` instead of `callId` to be consistent?

Comment (chubberlisk) [reply pages/wards/visits/[id]/edit.js:56]: Could I do dis in separate PRs after merging this one in?

Comment (jivdhaliwal) [reply pages/wards/visits/[id]/edit.js:56]: yep agreed

PR #540, opened 2020-07-23
Add fork sync github workflow

jivdhaliwal:
# What
Adds [fork sync](https://github.com/TG908/fork-sync) as a github workflow.

# Why
https://github.com/wei/pull relies on an external service which as been down for a while and is less reliable then using github actions

# Screenshots

# Notes
The workflow has been setup to explicitly declare which repos it will run on

### DIFF
--- a/.github/workflows/fork-sync.yml
+++ b/.github/workflows/fork-sync.yml
@@ -1,0 +1,18 @@
+name: Fork Sync
+
+on:
+  schedule:
+    - cron: "*/10 * * * *"
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/fork-sync.yml
```diff
@@ -0,0 +1,19 @@
+name: Fork Sync
+
+on:
+  schedule:
+    - cron: "*/10 * * * *"
```
chubberlisk, line 5: What does this translate to?

jivdhaliwal: Every 10 minutes

PR #541, opened 2020-07-23
Fix not sending updated visit text when mobile number is added

chubberlisk:
# What

- Fix `TypeError` showing in terminal when no edit a visit notification.
- Fix not sending updated visit text when mobile number is added with an already existing email address.

# Why

We want to ensure that a text message is sent when a mobile number is added when updating the details of a booked visit.

# Screenshots

image

# Notes

N/A

### DIFF
--- a/pages/api/update-a-visit.js
+++ b/pages/api/update-a-visit.js
@@ -106,7 +106,7 @@ export default withContainer(
 
         if (!notificationType)
           return {
-            success: null,
+            success: true,
             errors: null,
           };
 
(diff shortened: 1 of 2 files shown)


pages/api/update-a-visit.js
```diff
@@ -104,7 +104,11 @@ export default withContainer(
             : scheduledCall.recipientNumber
         );
 
-        if (!notificationType) return;
+        if (!notificationType)
+          return {
+            success: null,
```
jivdhaliwal, line 109: It feels a bit weird to use null here to imply success below (checking `emailSuccess === false` instead of `!emailSuccess`).  If `notificationType` is false, that means that the callTime and the contactMethod have not changed. As such should we return `success: true` instead of `success: null` ?

PR #545, opened 2020-07-23
Update visit retrieving usecases to include complete visits that have not been cleared

jivdhaliwal:
We eventually want to be able to mark visits as complete but still show them in the list screen. To do that we need to be able to differentiate between visits that have been cleared and which have not.

This change updates the visit retrieving usecases to check for complete visits where pii_cleared_at IS NULL, which signifies that they are safe to display to the user

# Screenshots

# Notes

### DIFF
--- a/db/scripts/cleanup_scheduled_calls.contractTest.js
+++ b/db/scripts/cleanup_scheduled_calls.contractTest.js
@@ -105,5 +105,11 @@ describe("test cleanup script", () => {
     // archived visit cannot be retrieved
     const archivedVisit = await container.getRetrieveVisitByCallId()("4");
     expect(archivedVisit.scheduledCall).toBeNull();
+
+    // Only the future visit is retrieved when retrieving all visits
+    const { scheduledCalls: visits } = await container.getRetrieveVisits()({
+      wardId,
+    });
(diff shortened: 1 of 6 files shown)


jivdhaliwal:
> Could we update the contract tests for the usecases to cover the addition checks for `pii_cleared_at` and `COMPLETE` status?

As we don't have usecases to set the status to complete yet, I decided to use the cleanup script contract test to test the usecases (see 3fc40877d5c477793f28a568c18c72f4b5b71fd7). I just need to update it with the retrieveVisits usecase but I think that should be good enough to get coverage on the pii_cleared_at field the status changes. wdyt?

chubberlisk:
> > Could we update the contract tests for the usecases to cover the addition checks for `pii_cleared_at` and `COMPLETE` status?
> 
> As we don't have usecases to set the status to complete yet, I decided to use the cleanup script contract test to test the usecases (see [3fc4087](https://github.com/madetech/nhs-virtual-visit/commit/3fc40877d5c477793f28a568c18c72f4b5b71fd7)). I just need to update it with the retrieveVisits usecase but I think that should be good enough to get coverage on the pii_cleared_at field the status changes. wdyt?

Ah right, yeah I see! Dis works for me. :shipit:

PR #562, opened 2020-07-31
Refresh Trust Admin token expiry

jivdhaliwal:
# What
Regenerated the Trust Admin token within a refresh timeframe

# Why
So that users are not logged out if they are using the app

# Screenshots

# Notes
See https://github.com/madetech/nhs-virtual-visit/pull/352 for the original implementation of the token regeneration

### DIFF
--- a/src/usecases/verifyTrustAdminToken.test.js
+++ b/src/usecases/verifyTrustAdminToken.test.js
@@ -1,15 +1,16 @@
 import verifyTrustAdminToken from "./verifyTrustAdminToken";
 
 describe("verifyTrustAdminToken", () => {
+  const res = {
+    writeHead: jest.fn(() => res),
+    end: jest.fn(),
+    setHeader: jest.fn(),
+  };
+
   const req = {
     headers: {
       cookie: "token=sample.token.value",
     },
-  };
-
-  const res = {
-    writeHead: jest.fn(() => res),
-    end: jest.fn(),
   };
 
   it("calls the supplied callback if the token is valid", async () => {
@@ -23,6 +24,7 @@ describe("verifyTrustAdminToken", () => {
     };
     const container = {
       getTokenProvider: () => tokenProvider,
+      getRegenerateToken: () => jest.fn().mockReturnValue({}),
     };
 
     const result = verifyTrustAdminToken(callback)({ req, res, container });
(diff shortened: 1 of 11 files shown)


src/usecases/verifyTrustAdminToken.test.js
```diff
@@ -65,4 +67,55 @@ describe("verifyTrustAdminToken", () => {
       Location: "/trust-admin/login",
     });
   });
+
+  it("updates the cookie if the token is regenerated", () => {
+    const regenRes = {
+      writeHead: jest.fn(() => regenRes),
```
chubberlisk, line 73: I don't quite understand this. We create a response `regenRes` with a method `writeHead` with a stubbed response to return itself? 🤔

src/usecases/verifyTrustAdminToken.test.js
```diff
@@ -65,4 +67,55 @@ describe("verifyTrustAdminToken", () => {
...
+      "token=encodedToken; httpOnly; path=/;",
+    ]);
+
+    expect(regenRes.setHeader).toHaveBeenCalled();
+
+    expect(callback).toHaveBeenCalledWith(
+      expect.objectContaining({
+        authenticationToken: regeneratedToken,
+      })
+    );
```
chubberlisk, line 119: Do you think it's worth adding an expectation to check that the `regenerateToken` usecase has been called with the original `authenticationToken`?

src/usecases/verifyTrustAdminToken.test.js
jivdhaliwal: It's to support the following chain at the end of the usecase:
```
 res.writeHead(302, { Location: "/admin/login" }).end();
```
We call writeHead on the res object and then call end() on the res returned by writeHead. So to simulate that the writeHead mock needs to return the res.

chubberlisk: Ah I see that now, thanks!

src/usecases/verifyTrustAdminToken.test.js
```diff
@@ -65,4 +67,56 @@ describe("verifyTrustAdminToken", () => {
...
+      getRegenerateToken: () => regenerateTokenSpy,
+    };
+
+    verifyTrustAdminToken(callback)({
+      req: regenReq,
+      res: regenRes,
+      container,
+    });
+
+    expect(regenerateTokenSpy).toHaveBeenCalledWith(authenticationToken);
```
chubberlisk, line 110: 🙇‍♀️

PR #563, opened 2020-07-31
Regenerate WardStaff token

jivdhaliwal:
# What
Regenerates the authentication token for WardStaff using the regenerateToken usecase when it is about to expire

# Why
Ensures that the ward staff token does not expire if the WardStaff is actively using the app.

# Screenshots

# Notes
See https://github.com/madetech/nhs-virtual-visit/pull/352 for original token regeneration work

### DIFF
--- a/src/usecases/verifyToken.test.js
+++ b/src/usecases/verifyToken.test.js
@@ -23,6 +23,7 @@ describe("verifyToken", () => {
     };
     const container = {
       getTokenProvider: () => tokenProvider,
+      getRegenerateToken: () => jest.fn().mockReturnValue({}),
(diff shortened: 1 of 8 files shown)


src/usecases/verifyToken.test.js
```diff
@@ -68,4 +69,58 @@ describe("verifyToken", () => {
       Location: "/wards/login",
     });
   });
+
+  fit("updates the cookie if the token is regenerated", async () => {
```
chubberlisk, line 73:
```suggestion
  it("updates the cookie if the token is regenerated", async () => {
```

jivdhaliwal: eagle eye save 🙏

PR #586, opened 2020-08-11
Updated all Router.push and Link tags to use the correct syntax

Shaunwild97:
# What

Modified every use for Router.push and Link that was incorrectly used, to the correct syntax

# Why

Every page was throwing a 404 while navigating. This was preventing the Link tags from working as expected

### DIFF
--- a/src/components/TrustsTable/index.js
+++ b/src/components/TrustsTable/index.js
@@ -40,9 +40,15 @@ const TrustsTable = ({ trusts }) => (
                 }
               </td>
               <td className="nhsuk-table__cell">
-                <AnchorLink href={`/admin/trusts/${trust.id}/edit`}>
+                <AnchorLink
+                  href="/admin/trusts/[id]/edit"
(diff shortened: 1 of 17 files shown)


src/components/TrustsTable/index.js
```diff
@@ -40,9 +40,15 @@ const TrustsTable = ({ trusts }) => (
...
+                <AnchorLink
+                  href="/admin/trusts/[id]/edit"
+                  as={`/admin/trusts/${trust.id}/edit`}
+                >
                   Edit
-                  <span className="nhsuk-u-visually-hidden"> {trust.name}</span>
+                  <span className="nhsuk-u-visually -hidden">
+                    {" "}
+                    {trust.name}
+                  </span>
```
chubberlisk, line 51: Looks like a space was added by accident?

chubberlisk:
Add yourself to our [Contributors](https://github.com/madetech/nhs-virtual-visit#contributors) list!

PR #632, opened 2020-10-08
speprate wardId from visit

maysakanoni:
# What
This commit separates `wardId` from `visit` in `createVisit`
# Why
Some time was spent refactoring `createVisit` so it accepts `wardId` and `trustId` as separate arguments because they are distinct domains from a `visit`. However the create visit function then merges these two concepts together and a `wardId` becomes part of a `visit`. This commit separates them.

### DIFF
--- a/src/usecases/createVisit.js
+++ b/src/usecases/createVisit.js
@@ -17,20 +17,19 @@ const createVisit = (
   if (!wardId) throw "creating visit with no wardId";
   if (!trustId) throw "creating visit with no trustId";
 
-  const newVisit = Object.assign(visit, { wardId, trustId });
-  const { validVisit, errors } = validateVisit(newVisit);
+  const { validVisit, errors } = validateVisit(visit);
 
   if (!validVisit) {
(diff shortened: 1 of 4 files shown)


src/usecases/createVisit.js
```diff
@@ -42,19 +41,15 @@ const createVisit = (
...
     callPassword,
     provider: trust.videoProvider,
-    wardId: visit.wardId,
   });
 
   try {
     return await db.tx(async (t) => {
       logger.debug("inserting visit");
-      await insertVisitQuery(t, populatedVisit);
+      // const { id, call_id } = await insertVisit(t, visit);
```
rebeccafitzsimmons1, line 49: I think we can remove this line too?

rebeccafitzsimmons1:
looks great! I'm happy that the `newVisit` logic in `createVisit` that we introduced as an intermediary step the other day whilst trying to fix the tests has gone too 😄

src/usecases/createVisit.js
maysakanoni: Yes removing this now

PR #633, opened 2020-10-08
Remove use of Real client in test and stub Fake client key

maysakanoni:
# What
Removed tests which were using the real Notify test client and remove the need for fake client to receive api key from the env var.
# Why
-Reduce the external dependencies on our tests
-Simplify the pipeline as an env var for the api key is not needed for the Fake client

### DIFF
--- a/src/gateways/GovNotify/index.js
+++ b/src/gateways/GovNotify/index.js
@@ -8,11 +8,10 @@ export default (() => {
   return {
     getInstance: async () => {
       if (!instance) {
-        const apiKey = process.env.API_KEY;
-
         if (process.env.APP_ENV === "test") {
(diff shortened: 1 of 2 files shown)


src/gateways/GovNotify/index.js
```diff
@@ -8,11 +8,10 @@ export default (() => {
   return {
     getInstance: async () => {
       if (!instance) {
-        const apiKey = process.env.API_KEY;
-
         if (process.env.APP_ENV === "test") {
-          instance = new FakeNotifyClient(apiKey);
+          instance = new FakeNotifyClient("apiKey");
```
faissalMT, line 12: Is this argument necessary?

PR #641, opened 2020-10-22
Fix the login redirect for the same ward and admin code

maysakanoni:
# What
 A guard clause is added to check if a password exists before a ward code can be verified and a token generated.
# Why
For a token to be generated, the code passed in to the session api must first be validated. `Trust admin` codes and `admin` codes require passwords in order for them to be verified as codes for which a token can be generated, `ward `codes do not. This means that if a `ward` code was the same as an `admin` code, when a login request is made using the `admin`, both the `admin` code and `ward` code is verified (as a password is provided for the `admin` verification and `ward` verification only requires for there to be a `ward` with that code) allowing a token to be generated for the `ward `code and set as the cookie. If a token exists for a `ward` code it would automatically redirect to the ward login page.

### DIFF
--- a/cypress/integration/admin/loggingInAndOut.js
+++ b/cypress/integration/admin/loggingInAndOut.js
@@ -31,12 +31,31 @@ describe("As an admin, I want to log in so that I can access the service.", () =
     ThenISeeAnError();
   });
 
+  it("allows an admin to log in and out when a ward code and admin code are the same", () => {
+    GivenTheSameWardCodeAsAdminCode();
+    GivenIAmAnAdmin();
+    WhenIVisitTheAdminLogInPage();
+
+    cy.audit();
+
+    AndIEnterAValidAdminCodeAndPassword();
+    AndISubmitTheForm();
+    ThenISeeTheAdminHomePage();
+
+    cy.audit();
+
+    thenIClickLogOut();
+    ThenISeeTheAdminLogInPage();
+  });
+
   it("displays an error for an invalid password", () => {
     WhenIVisitTheAdminLogInPage();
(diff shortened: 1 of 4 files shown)


cypress.json
```diff
@@ -8,7 +8,8 @@
     "validAdminCode": "super",
     "validAdminPassword": "adminpassword",
     "fakeWard": "FAKENHS",
-    "baseUrl": "http://localhost:3001"
+    "baseUrl": "http://localhost:3001",
+    "ValidDupeWard": "super"
```
rebeccafitzsimmons1, line 12: is this cypress env var being used anywhere?

package.json
```diff
@@ -25,6 +25,7 @@
     "start": "NODE_ENV=production node server.js",
     "cleandb": "node db/scripts/cleanup_scheduled_calls.js",
     "db:seed": "node db/scripts/seed_database.js",
+    "db:dupeseed": "node db/scripts/seed_same_ward_code_as_admin_code.js",
```
rebeccafitzsimmons1, line 28: might be worth having a slightly more descriptive name for this command?

rebeccafitzsimmons1 (commented):
This looks really good to me! Would it be worth having a cypress test (if there isn't one already) for the ward side of things too, to check that the user is still able to log in as a ward if the ward code is the same as an admin code?

package.json
maysakanoni: Yeah I agree this wasn't very clear. This has now been removed and into the single migration script for the test

cypress.json
maysakanoni: no, its been removed

PR #642, opened 2020-10-22
Refactor dependencies in Create Visit flow

rebeccafitzsimmons1:
# What
- Added a CreateVisitContainer to import the necessary dependencies into the `book-a-visit` controller.

- Pulled the ward and trust retrieval, and the getting of the callId and callPassword, up from the `createVisit` usecase to the controller. Also moved the retrieve ward and trust usecases to the gateway layer, as they make db calls and have no business logic.

- Get db instance in the gateways rather than from AppContainer.

- Added tests to check the db transaction in the `createVisitUnitOfWork` rolls back the db calls on error.

# Why

So that we can start moving away from needing AppContainer in the `book-a-visit` controller and just have it import the dependencies that it needs (though `book-a-visit` still needs access to AppContainer for now due to the `userIsAuthenticated` dependency).

Allows the usecase to not be concerned with ward and trust retrieval and getting call ids and passwords, and just focus on creating a visit.

# Notes
Future work could look towards refactoring the `userIsAuthenticated` out of `book-a-visit` controller, maybe into its own auth container? And could also look at pulling the `sendBookingNotification` construction out of the controller too (though bear in mind that `update-a-visit` also has a dependency on `sendBookingNotification` via the AppContainer).


src/gateways/UnitsOfWork/createVisitUnitOfWork.contractTest.js
```diff
@@ -89,7 +93,67 @@ describe("createVisitUnitOfWork tests", () => {
       ward
     );
 
-    expect(bookingNotificationSuccess).toEqual(true);
-    expect(bookingNotificationErrors).toEqual(null);
+    expect(success).toEqual(true);
+    expect(error).toEqual(null);
+  });
+
+  it("rolls back db calls when sendBookingNotification throws error in transaction", async () => {
```
MrTortoise, line 100: yay!

src/usecases/createVisit.js
```diff
@@ -1,54 +1,34 @@
...
-  }
-
-  const callId = await getCallIdProvider(trust, visit.callTime);
-
-  const callPassword = getRandomIdProvider().generate(10);
   const populatedVisit = Object.assign({}, visit, {
     callId,
     callPassword,
-    provider: trust.videoProvider,
+    provider: videoProvider,
```
MrTortoise, line 21: props for this

src/usecases/createVisit.test.js
```diff
@@ -43,42 +29,40 @@ describe("createVisit", () => {
...
-      retrieveWardByIdStub,
-      createVisitUnitOfWorkSpy
-    )(visit, wardId, trustId);
+    const response = await createVisit(createVisitUnitOfWorkSpy)(
+      visit,
+      ward,
+      callId,
+      callPassword,
+      videoProvider
+    );
```
MrTortoise, line 38: this is why you are doing this. Feels much better now imo

MrTortoise: Not suggesting this for now ... but maybe it's possible to remove the partial application altogether here? Just pass the UOW as an argument?

rebeccafitzsimmons1: yeah good point. I remember you mentioning this last week too. This point makes me think about the way we are currying functions in general - as right now it seems there's a pattern of passing the function's dependencies through in the first argument set, treating that as a constructor, and then passing the function args through in second argument set. I suppose now that we are starting to move away from the AppContainer and are importing dependencies when we need them, do we still have the same need for curried functions? - e.g. in `book-a-visit` we now pass the unitOfWork to `createVisit` at the same time as we pass all the other args, so are in a place to be able to pass through all the args at the same time.

PR #644, opened 2020-10-26
Add a feature flag provider

rebeccafitzsimmons1:
# What
Adds a feature flag provider.

# Why
So that we can deploy features without releasing them, and are able to toggle features on per environment.

# Notes
The feature flag provider takes an argument that corresponds to the name of the feature flag environment variable for the feature, e.g:

```
if (featureIsEnabled(`FEATURE_FLAG_TEST_FEATURE`)) {
    .....
}
```

where `featureIsEnabled` is the feature flag provider and, for the sake of this example, `FEATURE_FLAG_TEST_FEATURE` is the name of the env var for the 'test feature' feature flag. The feature flag env var needs to be set to `true` for the feature to be toggled on, e.g.

`FEATURE_FLAG_TEST_FEATURE=true`

and toggled to false, or simply omitted as an env var altogether, if the feature is to be toggled off.

### DIFF
--- a/src/helpers/featureFlag.js
+++ b/src/helpers/featureFlag.js
@@ -1,5 +1,5 @@
 const featureIsEnabled = (feature) => {
-  return process.env[feature] === "yes" ? true : false;
+  return process.env[feature] === "true";
 };
 
 export default featureIsEnabled;
--- a/src/helpers/featureFlag.js
+++ b/src/helpers/featureFlag.js
@@ -1,0 +1,5 @@
+const featureIsEnabled = (feature) => {
+  return process.env[feature] === "yes" ? true : false;
+};
+
+export default featureIsEnabled;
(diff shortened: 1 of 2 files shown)


src/helpers/featureFlag.js
```diff
@@ -0,0 +1,5 @@
+const featureIsEnabled = (feature) => {
+  return process.env[feature] === "yes" ? true : false;
```
MrTortoise, line 2: maybe "yes" and no could be true and false?

MrTortoise (commented):
I think true and false is better ... and yeah that makes it even less 👍

src/helpers/featureFlag.js
faissalMT, line 2:
```suggestion
  return process.env[feature] === "yes";
```
`===` already produces a boolean so the `?` is unnecessary

rebeccafitzsimmons1: lool monday brain 🤦‍♀️

PR #649, opened 2020-11-02
Add correlation id and session id

maysakanoni:
# What
-Add a session id to the cookies
-Add a correlation Id to headers
-Extract fetch request with correlation id into method
# Why
-This is some of the data and metadata required to log an event
# Notes
- A `console.log(correlationId)` has been added as a placeholder for where the correlation id will be needed when it cannot be passed as part of a fetch request to the backend. e.g. when a call ends, a fetch request api is not made but a correction id is generated as this event still needs to be logged

### DIFF
--- a/pages/admin/login.js
+++ b/pages/admin/login.js
@@ -1,5 +1,4 @@
 import React, { useRef, useState } from "react";
-import fetch from "isomorphic-unfetch";
 import Button from "../../src/components/Button";
 import ErrorSummary from "../../src/components/ErrorSummary";
 import FormGroup from "../../src/components/FormGroup";
@@ -10,8 +9,10 @@ import Label from "../../src/components/Label";
 import Layout from "../../src/components/Layout";
 import propsWithContainer from "../../src/middleware/propsWithContainer";
 import Form from "../../src/components/Form";
+import fetchEndpointWithCorrelationId from "../../src/helpers/fetchEndpointWithCorrelationId";
+import { v4 as uuidv4 } from "uuid";
 
-const Login = () => {
+const Login = ({ correlationId }) => {
   const [errors, setErrors] = useState([]);
 
   const codeRef = useRef();
@@ -46,13 +47,13 @@ const Login = () => {
     }
 
     if (onSubmitErrors.length === 0) {
-      const response = await fetch("/api/session", {
-        method: "POST",
-        body: JSON.stringify({ code, password }),
-        headers: {
-          "content-type": "application/json",
-        },
-      });
+      const body = JSON.stringify({ code, password });
+      const response = await fetchEndpointWithCorrelationId(
+        "POST",
+        "/api/session",
+        body,
+        correlationId
+      );
 
       if (response.status === 201) {
         window.location.href = `/admin`;
@@ -127,6 +128,8 @@ export const getServerSideProps = propsWithContainer(
     const adminIsAuthenticated = container.getAdminIsAuthenticated();
     const adminToken = adminIsAuthenticated(headers.cookie);
 
+    const correlationId = `${uuidv4()}-admin-login`;
+
     if (trustAdminToken) {
       res.writeHead(307, { Location: `/trust-admin` }).end();
     } else if (userToken && userToken.ward) {
@@ -135,6 +138,6 @@ export const getServerSideProps = propsWithContainer(
       res.writeHead(307, { Location: `/admin` }).end();
     }
 
-    return { props: {} };
+    return { props: { correlationId } };
   }
 );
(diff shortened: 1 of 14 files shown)


pages/admin/login.js
```diff
@@ -10,8 +9,10 @@ import Label from "../../src/components/Label";
 import Layout from "../../src/components/Layout";
 import propsWithContainer from "../../src/middleware/propsWithContainer";
 import Form from "../../src/components/Form";
+import fetchEndpointWithCorrelationId from "../../src/helpers/fetchEndpointWithCorrelationId";
+import { v4 as uuidv4 } from "uuid";
```
faissalMT, line 13: I did not know you could do this, nice

pages/api/session.js
```diff
@@ -1,92 +1,100 @@
...
+        wardCode: undefined,
+        trustId: verifyTrustAdminCodeResponse.trust.id,
+        type: TRUST_ADMIN,
       });
-      res.end(JSON.stringify({ wardId: code }));
-      return;
+    } else if (
+      verifyWardCodeResponse.validWardCode &&
+      !verifyTrustAdminCodeResponse.validTrustAdminCode &&
+      !verifyAdminCodeResponse.validAdminCode
```
faissalMT, line 47: Instead of checking all these you can simply move this to an `else if` at the end of the chain because the other conditions will be implicitly false

src/helpers/fetchEndpointWithCorrelationId.test.js
```diff
@@ -0,0 +1,33 @@
...
+
+describe("fetchEndpointWithCorrelationId", () => {
+  let body = JSON.stringify({ code: "code" });
+  let correlationId = `${uuidv4()}-login`;
+  let endpoint = "/api/session";
+  let method = "POST";
+  let scope;
+
+  beforeEach(() => {
+    scope = nock("http://localhost:3001", {
```
faissalMT, line 13: We should probably have the target API set by an environment variable so we don't have to change it in multiple places

pages/api/session.js
maysakanoni: nice!

src/helpers/fetchEndpointWithCorrelationId.test.js
maysakanoni: yeah agreed, I think its only being used in one other place for now.

pages/visitors/[id]/name.js
```diff
@@ -14,8 +14,12 @@ import propsWithContainer from "../../../src/middleware/propsWithContainer";
 import Error from "next/error";
 import Form from "../../../src/components/Form";
 import logger from "../../../logger";
+import { v4 as uuidv4 } from "uuid";
 
-const Name = ({ callId, error, callPassword }) => {
+const Name = ({ callId, error, callPassword, correlationId }) => {
+  useEffect(() => {
```
Comment (rebeccafitzsimmons1) [inline pages/visitors/[id]/name.js:20]: I think my react is rusty but why do we need `useEffect` here?

src/helpers/fetchEndpointWithCorrelationId.test.js
MrTortoise: yeah i'd leave it as is ... when we want to change it then you have validated your hypothesis that it should be external, change it then.

src/helpers/fetchEndpointWithCorrelationId.test.js
```diff
@@ -0,0 +1,33 @@
...
+  });
+  it("can make a request using fetchEndpointWithCorrelationId", async () => {
+    const response = await fetchEndpointWithCorrelationId(
+      method,
+      "http://localhost:3001/api/session",
+      body,
+      correlationId
+    );
+    expect(response.status).toBe(201);
+    scope.isDone();
```
MrTortoise, line 31: do we have a test that shows the backend gets the correlation id?

pages/visitors/[id]/name.js
Comment (rebeccafitzsimmons1) [reply pages/visitors/[id]/name.js:20]: sorry plz ignore me ^ - it makes sense now that I have read the notes section of the PR description :)

Comment (maysakanoni) [reply pages/visitors/[id]/name.js:20]: As far as I know its best practice to wait until react has updated the dom when you dont have code that executes as a result of a specific event in the page (like on onlick) but instead happens at some point during the lifecycle of the component. I think this is how I see this, that once the call has been ended ( i.e when this component mounts, which I think is the equivalent of use effect in hooks?), then the event logger would be called.

src/helpers/fetchEndpointWithCorrelationId.test.js
maysakanoni: I think this test is implicitly testing that, so as long as this function is used, then the correlationId is being sent as part of the headers. Nock requires the request to be identical, including for all of the headers to be set and will only then respond with a 201. There was a test which tested this assumption but was removed because it was acting up remotely.  All the api tests are use stubs and spies for the request, so they dont really test them at the integration level and as there is nothing that consumes the correlation id, we cant really test it in this way.

PR #651, opened 2020-11-04
Log login event using azure function

maysakanoni:
# What
Add a logEvent gateway which posts a http request to an azure function url, this then stores the data in cosmos-db in azure
Call the gateway in login class, when a ward staff logs in.
# Why
This is the first step to recording the user journey events
# Notes
This is only recording a login event for ward staff and not admins or trust admins, this not part of the immediate logging journey we want to look at but something to do in the future.

### DIFF
--- a/pageTests/api/session.test.js
+++ b/pageTests/api/session.test.js
@@ -73,9 +73,7 @@ describe("api/session", () => {
           validTrustAdminCode: false,
         }));
         const tokenGeneratorSpy = jest.fn(() => "generatedToken");
-        const logEventSpy = jest.fn(async () => ({
-          logEventResponse: true,
-        }));
+
         const container = {
           getTokenProvider: jest.fn(() => ({
             generate: tokenGeneratorSpy,
@@ -84,7 +82,6 @@ describe("api/session", () => {
           getVerifyTrustAdminCode: () => verifyTrustAdminCodeSpy,
           getVerifyAdminCode: () =>
             jest.fn().mockReturnValue({ validAdminCode: false }),
-          getLogEventGateway: () => logEventSpy,
         };
 
         await session(validRequest, response, { container });
--- a/pageTests/api/session.test.js
+++ b/pageTests/api/session.test.js
@@ -106,7 +106,66 @@ describe("api/session", () => {
           })
         );
       });
+
       it("logs a login event when a user has logged in", async () => {
+        process.env.EVENT_LOGGING = true;
+
+        const validRequest = {
+          method: "POST",
+          body: {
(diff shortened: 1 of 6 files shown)


src/gateways/logEvent.js
```diff
@@ -0,0 +1,18 @@
...
+  const response = await fetch(url, {
+    method: "POST",
+    body: body,
+    headers: {
+      "x-functions-key": key,
+    },
+  }).catch((error) => {
+    console.error("Error (log event):", error);
+    console.log(`the url is ${url}`);
+  });
```
faissalMT, line 14: Use the `try {} catch (){}` pattern here if you can, it ends up being a bit cleaner than dealing with promise callbacks especially given you're already making use of `await` here rather than `.then()`

src/gateways/logEvent.test.js
```diff
@@ -0,0 +1,46 @@
...
+    expect(event).toHaveProperty("correlationId");
+    expect(event).toHaveProperty("createdOn");
+    expect(event).toHaveProperty("streamName");
+    expect(event).toHaveProperty("trustId");
+    expect(event).toHaveProperty("eventType");
+    expect(event).toHaveProperty("event");
+    expect(event.event).toHaveProperty("wardId");
+    scope.isDone();
+  });
+});
```
faissalMT, line 46: This is a nitpick but paragraphing between groups of code (e.g: to separate variable declarations from assertions or to distinguish between blocks) typically makes code easier to read

pageTests/api/session.test.js
```diff
@@ -103,6 +106,75 @@ describe("api/session", () => {
...
+        });
+        expect(response.writeHead).toHaveBeenCalledWith(
+          201,
+          expect.objectContaining({
+            "Set-Cookie": [
+              expect.stringContaining("generatedToken"),
+              expect.stringContaining("sessionId"),
+            ],
+          })
+        );
```
rebeccafitzsimmons1, line 163: do we need to be asserting on these spies as part of this test - are they not checked as part of other, existing tests?

pageTests/api/session.test.js
```diff
@@ -103,6 +106,75 @@ describe("api/session", () => {
           })
         );
       });
+      it("logs a login event when a user has logged in", async () => {
```
rebeccafitzsimmons1, line 109: it might be good to have a test around what happens when the `logEvent` gateway returns an error?

rebeccafitzsimmons1:
closing and re-opening as github is saying there's conflicts with master but git is saying there's not.

PR #656, opened 2020-11-11
Remove not null constraint from patient name and visitor name

rebeccafitzsimmons1:
# What
Remove the `NOT NULL` database constraint from the `patient_name` and `recipient_name` fields.

# Why
Seems that due to the cleanup script (which is set to run against the trusts' heroku databases), the `patient_name` and `recipient_name` fields will be set to null every 24 hours. Therefore the migrations that moves the patient and visitor details over to the the new tables will fail when such null data is moved across to fields with `NOT NULL` constraints.

# Notes
we need the new migration that removes the not null constraint for the case where the migration has already run successfully (e.g. the heroku db for our staging app), but we need the not null removing from the existing migrations for the case where the migration has failed (e.g. the heroku db for the LNW staging app) - there is no side effect of removing a not null constraint from a field that doesn't already have a not null constraint.


rebeccafitzsimmons1:
closing due to merge conflict mess caused by changelog file

PR #671, opened 2020-12-17
Remove Jitsi support

faissalMT:
# What
Remove Jitsi support as well as references to it. Also fixes E2E tests not making use of environment variables and some contract test flakiness.

# Why
Jitsi was not used historically and means we have to maintain more code, removing support for this allows us to simplify our code


miranda-hawkes:
E2E tests are failing on this sorry

PR #676, opened 2020-12-21
Vv 71 trust manager flow add hospital code and status, removed URLs

nax2uk:
# What
  Added hospital code and hospital status to relevant hospital pages and removed support and survey URL from UI.

# Why
   Updated User requirements
# Screenshots
image
image

# Notes

### DIFF
--- a/pages/trust-admin/hospitals/add.js
+++ b/pages/trust-admin/hospitals/add.js
@@ -31,7 +31,6 @@ const AddAHospital = ({ trust, error, trustId }) => {
   };
 
   const submit = async (payload) => {
-    console.log(payload);
     payload.trustId = trustId;
     try {
       const response = await fetch("/api/create-hospital", {
--- a/pages/trust-admin/hospitals/add.js
+++ b/pages/trust-admin/hospitals/add.js
@@ -31,6 +31,7 @@ const AddAHospital = ({ trust, error, trustId }) => {
   };
 
   const submit = async (payload) => {
+    console.log(payload);
     payload.trustId = trustId;
     try {
       const response = await fetch("/api/create-hospital", {
--- a/src/components/EditHospitalForm/index.js
+++ b/src/components/EditHospitalForm/index.js
@@ -8,6 +8,7 @@ import Form from "../Form";
 import isPresent from "../../helpers/isPresent";
 import SelectStatus from "../../components/SelectStatus";
 
+
 const EditHospitalForm = ({ errors, setErrors, hospital = {}, submit }) => {
   const [hospitalName, setHospitalName] = useState(hospital.name);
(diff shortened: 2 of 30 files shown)


pages/trust-admin/hospitals/add.js
```diff
@@ -24,12 +24,13 @@ const AddAHospital = ({ error, trustId }) => {
...
   const getGenericError = () => {
     return {
-      id: 'generic-error',
+      id: "generic-error",
       message: "Something went wrong, please try again later.",
     };
   };
 
   const submit = async (payload) => {
+    console.log(payload);
```
faissalMT, line 33:
```suggestion
```

src/components/EditHospitalForm/index.js
```diff
@@ -91,43 +87,48 @@ const EditHospitalForm = ({ errors, setErrors, hospital = {}, submit }) => {
...
+          <Select
+            id="hospital-status"
+            className="nhsuk-input--width-10 nhsuk-u-width-one-half"
+            prompt="Choose a hospital status"
+            options={[
+              { id: 1, name: "active" },
+              { id: 2, name: "disabled" },
+            ]}
+            onChange={(event) => {
+              setHospitalStatus(
```
faissalMT, line 122: Consider moving this out into its own function

nax2uk: I have created a SelectStatus component to handle hospital Status and ward Status (and other future statuses). Please review now. Thanks.

src/components/EditHospitalForm/index.js
```diff
@@ -5,17 +5,15 @@ import Heading from "../Heading";
...
-  const [hospitalSurveyUrl, setHospitalSurveyUrl] = useState(
-    hospital.surveyUrl
-  );
-  const [hospitalSupportUrl, setHospitalSupportUrl] = useState(
-    hospital.supportUrl
-  );
+  const [hospitalStatus, setHospitalStatus] = useState(hospital.status);
+  const [hospitalCode, setHospitalCode] = useState(hospital.code);
+
+  const action = hospital.id ? "Edit" : "Add";
```
faissalMT, line 16: It's best to avoid the ternary operator for things like this as it makes things unclear

nax2uk: This was actually the previous code, which I re-added to the top of the code block. I think they wanted to combined two actions (Edit and Add) in one. How should I change the code? Should I add an if statement? Please suggest. Thanks.

faissalMT: Yeah an `if` would be better

nax2uk: If statement added. Please review, thank you.

faissalMT (approved):

PR #678, opened 2020-12-21
Refactor test package structure

miranda-hawkes:
# What
Put all unit and contract test specs in their own folder structures

# Why
To make it consistent and easier to find matching test specs and prevent test specs cluttering up prod code directories


test/usecases/retrieveHospitalWardVisitTotals.js
```diff
@@ -0,0 +1,45 @@
+export default ({ getDb }) => async (hospitalId) => {
```
faissalMT, line 1: I presume this was moved to the wrong place?

miranda-hawkes: Correct, just removed it

PR #679, opened 2020-12-21
Move ward query into gateway

faissalMT:
# What
This is the first of what will be a series of PRs moving SQL queries out of usecases and into gateways, which will later be placed under their own folder. This PR will likely serve as the precedent for future changes of this nature.

# Why
In order to being coupled to the database we need to pull every query out of the usecases, this is important because we plan to move to MSSQL. In addition, it also just makes it more convenient to write code in future.

### DIFF
--- a/src/gateways/findWardByCode.js
+++ b/src/gateways/findWardByCode.js
@@ -1,0 +1,12 @@
+const recordToDomain = ({ id, code, trust_id }) => ({
+  wardId: id,
+  wardCode: code,
+  trustId: trust_id,
+});
+const findWardQuery = async (db, wardCode) =>
+  await db.any(`SELECT id, code, trust_id FROM wards WHERE code = $1 LIMIT 1`, [
+    wardCode,
+  ]);
+
+export default ({ getDb }) => async (wardCode) =>
+  recordToDomain((await findWardQuery(await getDb(), wardCode))[0]);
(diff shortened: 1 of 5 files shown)


src/gateways/findWardByCode.js
```diff
@@ -0,0 +1,17 @@
+const recordToDomain = ({ id, code, trust_id }) => ({
```
miranda-hawkes, line 1: Do you want to put this into its own pg package to save refactoring later when we want to differentiate between postgres and mssql? Or will it be done as a large PR with the rest of the already separated out gateways?

src/usecases/verifyWardCode.js
```diff
@@ -1,20 +1,14 @@
 import logger from "../../logger";
 
-const verifyWardCode = ({ getDb }) => async (wardCode) => {
-  const db = await getDb();
-
+const verifyWardCode = ({ getFindWardByCodeGateway }) => async (wardCode) => {
```
miranda-hawkes, line 3: More of a question (also slacked you this but might be useful to share answer if others were wondering). At the moment the gateway will always be the postgres one with those SQL queries - in the future when we have both a pg and mssql gateway for the same query how will the correct gateway be decided?

src/gateways/findWardByCode.js
faissalMT: We can leave this how it is for now and move it into a package if we feel it's useful later.

src/usecases/verifyWardCode.js
faissalMT: Either the app container will choose which gateway to use or we will replace these gateways entirely

miranda-hawkes (approved):

PR #683, opened 2020-12-22
Vv 82 change env property pg

pl-madetech:
# What
Include PG_ suffix on database.js, removing the default constraints when using DATABASE_URL property.
Re-organisation on the db-migrate folder, to accommodate mssql on the next PR. 

# Why
Prepare project to include new mssql module, requires changes on the current env variables.

# Screenshots

# Notes
Improved local start up docker scripts + e2e test database script.


db/legacy_migrations/000_create_scheduled_calls_table.sql
```diff
@@ -1,7 +0,0 @@
-CREATE TABLE scheduled_calls_table (
```
faissalMT, line 1: Given we're still using Postgres why are we removing the migrations? This would prevent us from being able to setup the database again if needed

faissalMT (approved):

pl-madetech:
PR reviewed and merging to master.

PR #693, opened 2021-01-04
fix: error for inserting PIN data

nax2uk:
# What
    Fixed seed file for inserting PIN data
# Why
    Seed file structure change so did not rebase correctly.
# Screenshots

# Notes


pl-madetech (approved):
Approved. Quick question, do we need to continue working on dbmigration scripts on postgres side? I feel that we can have more problems in the future, if we continue changing dbmigrate files for postgres. Can we move to MSSQL and leave this scripts behind?

nax2uk:
> Approved. Quick question, do we need to continue working on dbmigration scripts on postgres side? I feel that we can have more problems in the future, if we continue changing dbmigrate files for postgres. Can we move to MSSQL and leave this scripts behind?
This was a fix to the prev sprint. Going forward, it will be MSSQL.

PR #694, opened 2021-01-04
Vv 79 add trusts

Richard-Pentecost:
# What
Ability to add an organisation (trusts) to a list of available organisations (trusts)
Added a status, so that trusts are disabled on creation until someone enables them

# Why
To allow for new organisations to be added to available organisations 

# Screenshots

# Notes


pl-madetech (approved):
All good richard.

PR #754, opened 2021-01-14
Vv 143 manager api tests

nax2uk:
# What
   Added unit test for the following apis
- archive-manager
- update-a-manager-status

# Why
Because tests are very essential.
# Screenshots
N/A
# Notes
N/A


nax2uk:
Please see Vv 143 edit and archive manager api tests pull request instead. Thanks

PR #787, opened 2021-01-21
Vv 151 organisation heading

nax2uk:
# What
-  swapped out getTrustById usecase for getOrganisationById usecase (which calls getOrganisationById mssql gateway)
- updated unit tests accordingly
- added contract tests

# Why
   migration to mssql
# Screenshots

# Notes

- I have skipped the contract test for getOrganizationById postgres gateway as it calls createOrganization in factories which now creates organisation in an mssql db.


pl-madetech (approved):
It's a pretty big one, but from the core code all looks good. Let's merge it.

PR #803, opened 2021-01-29
Insert visit & related fixes

faissalMT:
# What
Changes needed for insertVisit to work, based on the VV-162-FacilitiesAndDepartmentsGW branch which has now been merged in. This does not yet integrate insertVisit into the usecase layer because I want to get this along with the related fixes into people's local branches asap.  

# Why
To fix issues that were making writing code harder and to help continue to migrate to MSSQL

# Screenshots

# Notes
The history is very messy, this is the best I can realistically do.  
I've removed the test that tested for created data as a result of migrations because these should be done with seeds and it would now fail as a result of table truncation between tests.  
I've refactored some behaviour that's no longer needed as a result of tables now being truncated between test runs.  
visitStatus now includes functions for converting between ints and the constants as an interim solution


src/gateways/MsSQL/insertVisit.js
```diff
@@ -0,0 +1,26 @@
...
+        output inserted.id
+        values (@patient_name, @call_time, @recipient_name, @recipient_number, @recipient_email, @department_id, @status)`
+    );
+
+  return {
+    id: result.recordset[0].id,
+  };
+};
+
+export default insertVisit;
```
Richard-Pentecost, line 26: Should the db call have a try catch?

PR #808, opened 2021-02-01
chore: added retrieveDepartmentByCode GW, contract test, added GW to …

nax2uk:
…appContainer and appContainer test

# What

- converted findWardByCode to retrieveDepartmentByCode GW
- added contract test for GW

# Why
mssql migrations
# Screenshots

# Notes

PR #809, opened 2021-02-01
Vv 198 verify admin code

Richard-Pentecost:
# What

# Why

# Screenshots

# Notes


contractTest/usecases/verifyAdminCode.contractTest.js
```diff
@@ -2,7 +2,7 @@ import verifyAdminCode from "../../src/usecases/verifyAdminCode";
 import AppContainer from "../../src/containers/AppContainer";
 import setupAdmin from "../../test/testUtils/setupAdmin";
 
-describe("verifyAdminCode contract tests", () => {
+xdescribe("verifyAdminCode contract tests", () => {
```
faissalMT, line 5: Is this intentional?

Richard-Pentecost: Yes, I've commented them out because I'm not sure they are required anymore. The usecase is unit tested, and these tests are causing errors because I think they were written when the usecase wasn't using a gateway. I didn't want to delete the file in case it is still required

faissalMT:
Should be good aside from the things I've also mentioned in https://github.com/madetech/nhs-virtual-visit/pull/808

PR #814, opened 2021-02-02
Make visits be archived when a department is archived

two-first-names:
# What

# Why

# Screenshots

# Notes


nax2uk (approved):
All good 🙂

PR #824, opened 2021-02-04
Vv 201 contract test for update manager status by UUID

nax2uk:
# What

# Why

# Screenshots

# Notes


Richard-Pentecost (approved):
Tests are all good. Worth noting that the actual gateway might need a little looking at. There's no try catch block. I know this isn't part of this PR.

nax2uk:
> Tests are all good. Worth noting that the actual gateway might need a little looking at. There's no try catch block. I know this isn't part of this PR.

I have implemented it such that rejected promises / any other errors thrown in the gateways are being caught in the outer catch block in the use cases. I hope this is okay.

PR #857, opened 2021-02-17
Everything needed to get it deployed

two-first-names:


Richard-Pentecost (approved):

Richard-Pentecost:
`setup_local_mssql_docker.sh` and `setup_local_mssql_docker.sh' both have db-migrate scripts in them that won't work with the new package.json. This won't be captured by the tests.

two-first-names:
> `setup_local_mssql_docker.sh` and `setup_local_mssql_docker.sh' both have db-migrate scripts in them that won't work with the new package.json. This won't be captured by the tests.

Thanks, fixed in those. We can probably later as part of removing postgres, make these scripts simpler to understand/not refer to mssql.

PR #858, opened 2021-02-18
fix: authorise user flow

nax2uk:
# What
Fix Manager Sign Up Flow for Authorise 2nd or more Managers
# Why
Flow Not working
# Screenshots

# Notes

### DIFF
--- a/pages/sign-up/index.js
+++ b/pages/sign-up/index.js
@@ -37,7 +37,7 @@ const SignUp = ({ organisations, error }) => {
     setOrganisation(selectedOrganisation);
   };
 
-  const verifyEmail = (email) => email.match(/@nhs\.co\.uk/);
+  const verifyEmail = (email) => email.match(/@madetech\.com/);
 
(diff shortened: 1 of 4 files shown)


pages/sign-up/index.js
```diff
@@ -37,7 +37,7 @@ const SignUp = ({ organisations, error }) => {
     setOrganisation(selectedOrganisation);
   };
 
-  const verifyEmail = (email) => email.match(/@nhs\.co\.uk/);
+  const verifyEmail = (email) => email.match(/@madetech\.com/);
```
two-first-names, line 40: Would it be possible for this to be configurable from the start?
Hard coding like this is going to lead to issues further down the line when we inevitably need to change it.

nax2uk: Sorry, I accidentally changed that line. It was previously hardcoded to nhs.co.uk. I am going to revert it to the old code now.

two-first-names (approved):

PR #861, opened 2021-02-18
Vv 260 check sign up email address

nax2uk:
# What
validate email that is used to sign up a manager
# Why

# Screenshots

# Notes


src/helpers/validateSignUpEmailAddress.js
```diff
@@ -0,0 +1,13 @@
+import { ALPHANUMERIC } from "./validateEmailAddress";
+
+const validateSignUpEmailAddress = (email) => {
+    const emailDomain = process.env.NEXT_PUBLIC_SIGN_UP_EMAIL_DOMAIN
+
+    const EMAIL_DOMAIN_REGEX_STRING = emailDomain.replace(',','|').toUpperCase();
+    const EMAIL_VALIDATION_STRING =
+    `^[${ALPHANUMERIC}.!#$%&'*+/=?^_\`{|}~-]` + 
+    `+@${EMAIL_DOMAIN_REGEX_STRING}`
```
two-first-names, line 9: Add `([A-Za-z0-9\-]\.)*` here, to capture subdomains. `test@trust.nhs.uk` is a possible email address.

src/helpers/validateSignUpEmailAddress.js
```diff
@@ -0,0 +1,13 @@
+import { ALPHANUMERIC } from "./validateEmailAddress";
+
+const validateSignUpEmailAddress = (email) => {
+    const emailDomain = process.env.NEXT_PUBLIC_SIGN_UP_EMAIL_DOMAIN
```
two-first-names, line 4: Can this just be `SIGN_UP_EMAIL_DOMAINS`, please, `NEXT_PUBLIC` isn't necessary.

src/helpers/validateSignUpEmailAddress.js
```diff
@@ -0,0 +1,13 @@
...
+
+    const EMAIL_DOMAIN_REGEX_STRING = emailDomain.replace(',','|').toUpperCase();
+    const EMAIL_VALIDATION_STRING =
+    `^[${ALPHANUMERIC}.!#$%&'*+/=?^_\`{|}~-]` + 
+    `+@${EMAIL_DOMAIN_REGEX_STRING}`
+    const EMAIL_VALIDATION_REGEX = new RegExp(EMAIL_VALIDATION_STRING);
```
two-first-names, line 10: There's no reason to create the regex each time we call this function, extract it so it is only ran once.

terraform/app_service.tf
```diff
@@ -58,5 +58,6 @@ resource "azurerm_app_service" "app_service" {
     "SENTRY_PROJECT"                    = var.sentry_project
     "SENTRY_AUTH_TOKEN"                 = var.sentry_auth_token
     "ENABLE_SENTRY"                     = "no"
+    "NEXT_PUBLIC_SIGN_UP_EMAIL_DOMAIN"  = var.valid_email_domains
```
two-first-names, line 61: Can this just be SIGN_UP_EMAIL_DOMAINS, please, NEXT_PUBLIC isn't necessary.

two-first-names (changes requested):

src/helpers/validateSignUpEmailAddress.js
```diff
@@ -0,0 +1,13 @@
...
+    return EMAIL_VALIDATION_REGEX.test(email.toUpperCase());
```
two-first-names, line 11: Don't use `.toUpperCase()`, on line above do `new RegExp(EMAIL_VALIDATION_STRING, 'i')`
This will probably mean you can just use a-z rather than a-zA-Z

two-first-names (changes requested):

src/helpers/validateSignUpEmailAddress.js
nax2uk: To access the environment variables from the browser, I had to prefix the env variables with NEXT_PUBLIC. Else it will be undefined. 
Let me know if there is a better way of doing this? TIA

two-first-names: Aha, I didn't know this. :)
That's fine in that case.

nax2uk: Yeah, it's quite cool. 🙃

two-first-names, line 4: To deal with the possibility the env var isn't set, can you just add `|| ".+"` here, so any email will be allowed locally.

src/helpers/validateSignUpEmailAddress.js
```diff
@@ -0,0 +1,13 @@
+import { ALPHANUMERIC } from "./validateEmailAddress";
+
+const validateSignUpEmailAddress = (email) => {
+    const emailDomain = process.env.NEXT_PUBLIC_SIGN_UP_EMAIL_DOMAIN
+
+    const EMAIL_DOMAIN_REGEX_STRING = emailDomain.replace(',','|').toUpperCase();
+    const EMAIL_VALIDATION_STRING =
+    `^[${ALPHANUMERIC}.!#$%&'*+/=?^_\`{|}~-]` +
```
two-first-names, line 8: Escape the `-` and `^` here.

two-first-names (changes requested):

src/helpers/validateSignUpEmailAddress.js
nax2uk: Oh yeah, sorry I put them in the method when I coded it with two sign up env variables and env checking. Taking them out now. 🙂

terraform/app_service.tf
nax2uk: To access the environment variables from the front end / browser, I had to prefix the env variables with NEXT_PUBLIC. Else it will be undefined.  
Let me know if there is a better way of doing this? TIA

src/helpers/validateSignUpEmailAddress.js
nax2uk: does `const emailDomain = process.env.NEXT_PUBLIC_SIGN_UP_EMAIL_DOMAIN` go in the validateSignUpEmailAddress method or remain in it?

src/helpers/validateSignUpEmailAddress.js
nax2uk: I am not sure why my test hangs when I make these changes locally.
I am going to push it up to see if the test does not hang.

nax2uk: so, the test does not hang when pushed up. I am not sure if this is only me having a hanging problem for validateSignUpEmailAddress.test.js
Something to take note.

src/helpers/validateSignUpEmailAddress.js
nax2uk: Do we have to deal with the env var not being set locally. Shouldn't it be set?

I noticed that when `const emailDomain = process.env.NEXT_PUBLIC_SIGN_UP_EMAIL_DOMAIN` is declared outside the validateSignUpEmailAddress method, `process.env.NEXT_PUBLIC_SIGN_UP_EMAIL_DOMAIN` is `undefined` remotely. Thus making the unit test fail.
Why is this so and adding ".+" would mean that the unit test for validateSignUpEmailAddress will give false positive results?
Well this is my understanding but tell me what you think.

src/helpers/validateSignUpEmailAddress.js
nax2uk: ok. As discussed on slack. Regex now goes inside.

src/helpers/validateSignUpEmailAddress.js
nax2uk: As discussed on slack `emailDomain = process.env.NEXT_PUBLIC_SIGN_UP_EMAIL_DOMAIN` is now inside validateSignUpEmailAddress, and adding `|| ".+"`

src/helpers/validateSignUpEmailAddress.js
nax2uk: My test still fails locally, have to debug this issue on my end. 🙁 

So this issue is resolved.

two-first-names (approved):

PR #902, opened 2021-03-11
fix: bug in wards/visit-start page

nax2uk:
# What

# Why

# Screenshots

# Notes


Richard-Pentecost (commented):
What is the purpose of the `await setUpScheduledCall({ departmentId })` added in the contract test. It doesn't appear to do anything in the test

nax2uk:
> What is the purpose of the `await setUpScheduledCall({ departmentId })` added in the contract test. It doesn't appear to do anything in the test

This is to add more than one visit in the `scheduled_call` table to check if the correct visit is retrieved. If you remember the bug was that it retrieved the wrong visit because there was more than one visit.

Richard-Pentecost (approved):

PR #974, opened 2021-08-09
Investigation changes

MichealWilkinson-MT:


MichealWilkinson-MT:
Cannot get the contract and E2E tests running in this environment, I have simulated the environment locally IE same node version etc however we're not seeing these errors locally

PR #978, opened 2021-08-24
Fix lint issues

simon-mt:
# What
Corrected code which generated lint errors and warnings.

# Why
Neither warnings nor errors should be generated.

### DIFF
--- a/pages/admin/trusts/add-a-trust.js
+++ b/pages/admin/trusts/add-a-trust.js
@@ -40,14 +40,14 @@ const AddATrust = () => {
     }
 
     if (onSubmitErrors.length === 0) {
-      const submitTrust = async (trustName) => {
+      const submitTrust = async () => {
         const response = await fetch("/api/create-organisation", {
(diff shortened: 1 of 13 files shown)


pages/admin/trusts/add-a-trust.js
```diff
@@ -40,7 +40,7 @@ const AddATrust = () => {
     }
 
     if (onSubmitErrors.length === 0) {
-      const submitTrust = async (name) => {
+      const submitTrust = async () => {
```
LewisDaleUK, line 43: Removing the `name` argument from this function will cause the response body to be `undefined` - is this the right behaviour? Line 50 uses the parameter

simon-mt: Yeah, 'name' is defined as a state variable as at higher scope level so passing it through as a parameter to the function is unnecessary (also caused a lint warning).

LewisDaleUK (approved):

PR #981, opened 2021-08-27
Run all unit tests

simon-mt:
# What
Determine why the skipped unit tests have been set to skip and refactor them where necessary to ensure that all unit tests run.
# Why
To ensure we get as much value as possible from running the unit test suites.


test/usecases/retrieveAverageVisitsPerDayByTrustId.test.js
```diff
@@ -53,14 +56,17 @@ describe.skip("retrieveAverageVisitsPerDay", () => {
...
+      // averageVisitsPerDay,
       error,
     } = await retrieveAverageVisitsPerDayByTrustId(container)(
       trustId,
       new Date(2020, 6, 3)
     );
 
-    expect(averageVisitsPerDay).toEqual(2);
+    // TODO: Uncomment the following line once we understand fully why retrieveAverageVisitsPerDay returns a hardcoded value.
+    // expect(averageVisitsPerDay).toEqual(2);
```
LewisDaleUK, line 68: Looking at the original code, I think this is because the gateway being used to actually retrieve the number of average visits per day is just being mocked out and injected to the use case - so the test hardcodes the value so we can ensure that the use case is correctly interacting with the gateway.

It's not a hugely _valuable_ test in this case, it needs triangulation (maybe randomise the return value as a first pass?). But i think it can be added back in

simon-mt: In this example, the container used for injection is ignored by the use case - so it would never use a real or a mocked out gateway implementation (src/usecases/retrieveAverageVisitsPerDayByTrustedId.js)

LewisDaleUK: So it is! My mistake, sorry

LewisDaleUK (approved):

PR #997, opened 2021-09-14
Disable scrolling on date input

LewisDaleUK:
# What

Refactor date entry fields so that the resultant HTML matches the example used in: https://service-manual.nhs.uk/design-system/components/date-input

This prevents erroneous input, including those caused by accidentally scrolling the mousewheel

# Why
The design system sets out best practices that have been user tested quite thoroughly - we should only deviate from it wnen absolutely necessary and when we have performed our own testing

# Screenshots

# Notes

See: https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/

### DIFF
--- a/src/components/DateSelect/index.js
+++ b/src/components/DateSelect/index.js
@@ -48,7 +48,7 @@ const DateSelect = ({
           <FormGroup>
             <Label htmlFor="day">Day</Label>
             <Input
-              type="number"
+              type="text"
               hasError={hasDateError}
               onChange={(event) => {
                 setDate({ ...date, day: parseInt(event.target.value) });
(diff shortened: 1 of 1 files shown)


simon-mt (changes requested):
There is a typo on line 115 ("tet").

src/components/DateSelect/index.js
```diff
@@ -106,7 +112,7 @@ const DateSelect = ({
           <FormGroup>
             <Label htmlFor="hour">Hour</Label>
             <Input
-              type="number"
+              type="tet"
```
simon-mt, line 115: Typo here

simon-mt (approved):

